PR #3, opened 2017-10-18
Add articles views

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2400)

e.g., 
image


article/templates/article/base.html
```diff
@@ -0,0 +1,17 @@
...
+
+{% block css %}
+    <link href="{% static 'styles/pages/article.css' %}" rel="stylesheet" type="text/css" />
+{% endblock %}
+
+{% block content %}
+
+<div class="container">
+    <article id="top">
+        {% include_markdown article.markdown_file %}
```
konradko: are we passing filepath / file object or text? If text then it should be `article.markdown_text` or sth like that

article/templatetags/tests/test_article_tags.py
```diff
@@ -0,0 +1,19 @@
+from django.template.exceptions import TemplateDoesNotExist
+import pytest
+
+from article.templatetags import article_tags
+
+
+def test_include_markdown_valid_template():
```
konradko: would be good to try to render all markdown files in this test, hopefully some exception would be raised if the syntax is invalid - if not can you add such

article/templates/article/base.html
richtier: it's file path. Will rename to `markdown_file_path`.

article/views.py
```diff
@@ -0,0 +1,181 @@
+from django.views.generic import TemplateView
+
+
+class BaseArticleView(TemplateView):
+    template_name = 'article/base.html'
+
+    def get_context_data(self, *args, **kwargs):
+        return super().get_context_data(
```
konradko: think you need to move the extra kwarg to the end - think this will fail if there will be any args (as using kwarg, args, kwargs)

article/templatetags/tests/test_article_tags.py
richtier: we do - in the `test_article_views.py`

ui/urls.py
```diff
@@ -17,4 +17,209 @@
         core.views.LandingPagelView.as_view(),
         name='landing-page',
     ),
+    url(
```
konradko: so much urls 👍 :-)

PR #5, opened 2017-10-19
Ed 2385 add triage form view

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2385)

wizard

image


core/templates/core/widgets/radio.html
```diff
@@ -0,0 +1,5 @@
+{% with id=widget.attrs.id %}<ul{% if id %} id="{{ id }}"{% endif %}{% if widget.attrs.class %} class="{{ widget.attrs.class }}"{% endif %}>{% for group, options, index in widget.optgroups %}{% if group %}
```
konradko: very unreadable, please pretty print

richtier: this is copied and pasted from django's codebase with a few change. keeping it similar structure to theirs makes diffing with their file easier

konradko: why would you keep diffing?

richtier: someone other than myself may want to understand the differences between this file and Django's file - e.g., during an upgrade to Django 1.2.0 when Django's template may change.

PR #6, opened 2017-10-19
Persist triage answers

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2387)


triage/helpers.py
```diff
@@ -0,0 +1,45 @@
...
+        if request.sso_user is None:
+            return SessionTriageAnswersManager(request)
+        return DatabaseTriageAnswersManager(request)
+
+
+class SessionTriageAnswersManager(BaseTriageAnswersManager):
+    SESSION_KEY = 'TRIAGE_ANSWERS'
+
+    def persist_answers(self, answers):
+        self.request.session[self.SESSION_KEY] = answers
```
konradko: needs `self.request.session.modified = True`, alternatively we can add `SESSION_SAVE_EVERY_REQUEST = True` to settings

richtier: I'll do `modified = True` as the latter is less explicit of why it's happening.

PR #12, opened 2017-10-20
ED-2430 Custom page view and url

sdonk:
Top 10 exporters to be implemented in another PR


article/tests/test_article_views.py
```diff
@@ -43,7 +43,7 @@
     ),
     (
         views.UnderstandYourCustomersCultureView,
-        reverse('understand-your-cutomers-culture'),
+        reverse('understand-your-customers-culture'),
```
sdonk: Not really related to this PR, just a typo fix

conftest.py
```diff
@@ -0,0 +1,28 @@
+from unittest.mock import patch
+
+import pytest
+
+from sso.utils import SSOUser
+
+
+@pytest.fixture
+def sso_user():
```
sdonk: If this is used as a local fixture elsewhere, it will need to be replaced with the global fixture in a future PR

PR #17, opened 2017-10-23
Custom page

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2391)
[this task](https://uktrade.atlassian.net/browse/ED-2432)

# regular exporter

image


ui/urls.py
```diff
@@ -277,7 +277,7 @@
     url(
         r'^custom$',
         triage.views.CustomPageView.as_view(),
-        name='custom-page'
+        name='custom-landing-page'
```
konradko: This is just custom page, as landing is the home page

article/structure.py
```diff
@@ -256,11 +256,11 @@
     GUIDANCE_OPERATIONS_AND_COMPLIANCE_ARTICLES,
     ALL_ARTICLES,
 ]
-ALL_GOUPS_DICT = {group.key: group for group in ALL_GROUPS}
+ALL_GROUPS_DICT = {group.key: group for group in ALL_GROUPS}
 
 
 def get_article_group(group_key):
-    return ALL_GOUPS_DICT.get(group_key) or ALL_ARTICLES
+    return ALL_GROUPS_DICT.get(group_key) or ALL_ARTICLES
```
konradko: There is a typo in the “intellectual” as well

core/views.py
```diff
@@ -1,5 +1,13 @@
 from django.views.generic import TemplateView
+from triage.helpers import TriageAnswersManager
 
 
 class LandingPagelView(TemplateView):
     template_name = 'core/landing-page.html'
+
+    def get_context_data(self, *args, **kwargs):
+        answer_manager = TriageAnswersManager(self.request)
+        has_completed_triage = answer_manager.retrieve_answers() != {}
```
konradko: maybe it's worth adding it as a context processor? Cause that info will be needed on multiple pages

triage/forms.py
```diff
@@ -109,12 +109,20 @@ def get_persona(cleaned_data):
...
 
-def get_has_exported_before(cleaned_data):
-    return cleaned_data.get('exported_before') is True
+def get_has_exported_before(answers):
+    return answers.get('exported_before') is True
 
 
-def get_is_regular_exporter(cleaned_data):
-    return cleaned_data.get('regular_exporter') is True
+def get_is_regular_exporter(answers):
```
konradko: would probably be good to have these always in context?

triage/tests/test_triage_views.py
```diff
@@ -195,7 +196,7 @@ def test_companies_house_search_api_success(
 
 
 @patch('triage.helpers.DatabaseTriageAnswersManager.retrieve_answers')
-def test_custom_view(mocked_retrieve_answers, authed_client, sso_user):
+def test_custom_landing_page(mocked_retrieve_answers, authed_client, sso_user):
```
konradko: I'd still call it a custom page - users won't be landing on it.

triage/views.py
```diff
@@ -95,7 +96,46 @@ def dispatch(self, request, *args, **kwargs):
...
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        persona = forms.get_persona(self.triage_answers)
-        context['persona'] = persona
         context['triage_result'] = self.triage_answers
+        context['section_configuration'] = self.get_section_configuration()
         return context
+
+    def get_section_configuration(self):
+        persona = forms.get_persona(self.triage_answers)
```
konradko: feels like `get_persona` should be in utils/helpers

triage/forms.py
konradko: we could have these in the context processor as well

triage/views.py
```diff
@@ -95,7 +96,46 @@ def dispatch(self, request, *args, **kwargs):
...
+            return self.get_persona_new_section_configuration()
+        elif persona == forms.OCCASIONAL_EXPORTER:
+            return self.get_persona_occasional_section_configuration()
+        elif persona == forms.REGULAR_EXPORTER:
+            return self.get_persona_regular_section_configuration()
+
+    def get_persona_new_section_configuration(self):
+        return {
+            'persona_article_group': structure.PERSONA_NEW_ARTICLES,
+            'trade_profile': not forms.get_is_sole_trader(self.triage_answers),
```
konradko: IMHO methods without args/kwargs that are later accessing `self` are bit implicit, but I see the point here is to keep `get_section_configuration` cleaner

core/views.py
richtier: that would be quite implicit and really it's only used on two views

triage/views.py
richtier: `get_persona` needs to know the shape of the data generated by the forme - and it makes sense to keep it in one place in case field name changes then we can grep a single file to make the changes.

triage/forms.py
richtier: that would be increasing logic in the template. currently, thew view calls `get_is_regular_exporter` to help determine if certain sections of the custom template are to be rendered. We don't want to put logic more complex than `if...endif` in the template if we can help it.

triage/views.py
richtier: :+1:  I'll pass in self.triage_answers as a kwarg and turn these into static functions.

triage/views.py
konradko: `we can grep a single file to make the changes` that argument can be used to argue for keeping all of the code in one file :-P

IMHO just forms and things used by them should be in `forms.py`, while it seems to not be the case here.

ui/urls.py
```diff
@@ -8,13 +7,6 @@
 
 urlpatterns = [
     url(
-        r"^robots\.txt$",
```
konradko: Why delete?

richtier: robots.txt does not exist in this repo.

adding useful a robots.txt would require a sitemap.xml - overall adding useful robots.txt will take a few hours and I see no reason to block this PR for that feature.

Will add a ticket.

PR #18, opened 2017-10-23
ED-2401-record article visit

sdonk:


article/helpers.py
```diff
@@ -36,3 +40,49 @@ def build_linkedin_link(request, title):
...
+
+    def retrieve_articles(self):
+        return self.request.session.get(self.SESSION_KEY, [])
+
+
+class DatabaseArticlesReadManager(BaseArticleReadManager):
+
+    def persist_article(self, article):
+        response = api_client.exportreadiness.create_article_read(
+            form_data=article,
```
konradko: `form_data` on restful API endpoint? Why not just `article_uuid`?

article/views.py
```diff
@@ -7,6 +7,14 @@
 class BaseArticleDetailView(TemplateView):
     template_name = 'article/detail-base.html'
 
+    def get(self, request, *args, **kwargs):
+        self.record_article_visit()
+        return super().get(request, *args, **kwargs)
+
+    def record_article_visit(self):
+        manager = helpers.ArticleReadManager(self.request)
+        manager.persist_article(article={'article_uuid': self.article.uuid})
```
konradko: yeah defo better to just pass `article_uuid=self.article.uuid`

PR #20, opened 2017-10-24
Carousel

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2392)

image

image

image


casestudy/casestudies.py
```diff
@@ -0,0 +1,36 @@
+from collections import namedtuple
+
+
+from django.contrib.staticfiles.storage import staticfiles_storage
+
+
+Casestudy = namedtuple('Casestudy', ['title', 'description', 'image_url'])
```
konradko: According to https://stackoverflow.com/questions/36371047/python-naming-convention-namedtuples namedtuples have same naming convetion as classes, so `CaseStudy`

PR #21, opened 2017-10-24
Allow updating triage answers

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2446)


triage/views.py
```diff
@@ -52,6 +52,14 @@ class TriageWizardFormView(SessionWizardView):
     }
     success_url = reverse_lazy('custom-page')
 
+    @cached_property
```
konradko: Wonder if setting `self.triage_answers` in `__init__` would have the same result? Do we want to prevent `triage_answers` from being modified as well? Looks like you'd have to `del self.triage_answers` to be able to modify them https://docs.djangoproject.com/en/1.11/ref/utils/#module-django.utils.functional (if I understand correctly).

richtier: We want to retrieve from API/session only once.

We want to allow the user to update their answers-  so we combine the existing answers with the new answers when updating the answers.

note on page reload the value in `cached_property` is lost, so we won't have to worry about cleanup - we just want the data to be cached in the single request-response cycle.

PR #24, opened 2017-10-25
Add count method and tests

sdonk:


article/articles.py
```diff
@@ -1,42 +1,112 @@
 from collections import namedtuple
 
 from directory_constants.constants import exred_articles
-from django.core.urlresolvers import reverse_lazy
+from django.urls import reverse_lazy
```
sdonk: This is to silence the warning

article/helpers.py
```diff
@@ -56,3 +69,12 @@ def retrieve_articles(self):
         )
         response.raise_for_status()
         return [article['article_uuid'] for article in response.json()]
+
+    def article_read_count(self, category):
+        if category not in articles.CATEGORIES_LOOKUP:
+            raise Exception('Incorrect category')
```
sdonk: Don't like to raise a general `Exception` but this should never happen

article/articles.py
```diff
@@ -1,42 +1,112 @@
...
     uuid='GUIDANCE_MARKET_RESEARCH',
     title='Market research',
+    articles=frozenset((
+        exred_articles.DO_RESEARCH_FIRST,
+        exred_articles.DEFINE_MARKET_POTENTIAL,
+        exred_articles.DO_FIELD_RESEARCH,
+        exred_articles.ANALYSE_THE_COMPETITION,
+        exred_articles.VISIT_TRADE_SHOW
+    ))
 )
```
richtier: these are already defined in `structure.py` - see `GUIDANCE_MARKET_RESEARCH_ARTICLES` etc (the `ArticleGroup`).

Note `ArticleParent` is just for displaying this the name of the category the article is within e.g., "Market research".

image

I can see the naming is confusing. Perhaps we should instead remove `ArticleParent` do something like this:

```
CATEGORY_LABEL_MARKET_RESEARCH = 'Market research'

DO_RESEARCH_FIRST = Article(
    ...
    category_label=CATEGORY_LABEL_MARKET_RESEARCH,
)
DEFINE_MARKET_POTENTIAL = Article(
    ...
    category_label=CATEGORY_LABEL_MARKET_RESEARCH,
)
...


```

article/helpers.py
```diff
@@ -40,6 +43,19 @@ def persist_article(self, article_uuid):
     def retrieve_articles(self):
         return self.request.session.get(self.SESSION_KEY, [])
 
+    def article_read_count(self, category):
+        if category not in structure.CATEGORIES_LOOKUP:
+            raise Exception('Incorrect category')
```
konradko: A similar error will be raised by `CATEGORIES_LOOKUP[category]` - I would remove this unless we want to catch this specific error message for some reason and expect it to ever happen.

article/helpers.py
```diff
@@ -56,3 +72,15 @@ def retrieve_articles(self):
         )
         response.raise_for_status()
         return [article['article_uuid'] for article in response.json()]
+
+    def article_read_count(self, category):
+        if category not in structure.CATEGORIES_LOOKUP:
+            raise Exception('Incorrect category')
```
konradko: ditto

article/helpers.py
```diff
@@ -56,3 +72,15 @@ def retrieve_articles(self):
         )
         response.raise_for_status()
         return [article['article_uuid'] for article in response.json()]
+
+    def article_read_count(self, category):
+        if category not in structure.CATEGORIES_LOOKUP:
```
konradko: the read count is not just for guidance categories but for all possible article lists

article/helpers.py
```diff
@@ -56,3 +72,15 @@ def retrieve_articles(self):
...
+            raise Exception('Incorrect category')
+
+        read_articles = frozenset(self.retrieve_articles())
+        articles_in_category = frozenset(
+            [article.uuid for article in
+             structure.CATEGORIES_LOOKUP[category].articles]
+        )
+        read_articles_in_category = read_articles & articles_in_category
```
konradko: `read_articles_in_category = read_articles.intersection(articles_in_category)` would be much more explicit, I'd even consider adding a comment that `read_articles_in_category is a new set with elements common to read_articles and articles_in_category`

article/structure.py
```diff
@@ -194,6 +194,19 @@
...
+    GUIDANCE_MARKET_RESEARCH_ARTICLES,
+    GUIDANCE_CUSTOMER_INSIGHT_ARTICLES,
+    GUIDANCE_FINANCE_ARTICLES,
+    GUIDANCE_BUSINESS_PLANNING_ARTICLES,
+    GUIDANCE_GETTING_PAID_ARTICLES,
+    GUIDANCE_OPERATIONS_AND_COMPLIANCE_ARTICLES,
+]
+
+
+CATEGORIES_LOOKUP = {group.key: group for group in CATEGORIES_GROUPS}
```
konradko: `ALL_GROUPS` not just guidance categories

article/helpers.py
```diff
@@ -56,3 +72,15 @@ def retrieve_articles(self):
         )
         response.raise_for_status()
         return [article['article_uuid'] for article in response.json()]
+
+    def article_read_count(self, category):
```
konradko: `group` instead of `category`

article/helpers.py
```diff
@@ -40,6 +43,19 @@ def persist_article(self, article_uuid):
     def retrieve_articles(self):
         return self.request.session.get(self.SESSION_KEY, [])
 
+    def article_read_count(self, category):
```
konradko: this looks same as the method in the other class? Should be in `BaseArticleReadManager`

article/helpers.py
sdonk: This time I strongly don't agree. '&' is the intersection operator in Python, how `intersection()` would be more explicit? Then we should use `__add__` instead of `+`?

article/helpers.py
```diff
@@ -56,3 +72,15 @@ def retrieve_articles(self):
         )
         response.raise_for_status()
         return [article['article_uuid'] for article in response.json()]
+
+    def article_read_count(self, category):
+        if category not in structure.CATEGORIES_LOOKUP:
+            raise Exception('Incorrect category')
+
+        read_articles = frozenset(self.retrieve_articles())
+        articles_in_category = frozenset(
```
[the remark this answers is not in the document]
sdonk: Articles in `ArticleGroup` need to be a list because they have to be ordered. Turning a list into a frozenset is a very cheap operation, there are not very many articles (50?) and not very many users. Explicitly casting the list to a frozenset here, where the set is the intersected, makes the code more readable.

article/helpers.py
sdonk: I had this discussion with @richtier already.  Abstract classes cannot have concrete methods. Having a concrete base class would mean adding another level of inheritance, which seems unnecessary and a bit pointless when trying to save 3 lines of code. I'd rather turn the abstract class into a concrete class and move the common method there.

PR #30, opened 2017-10-26
Split sole trader and company into multiple questions

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2433)
[this task](https://uktrade.atlassian.net/browse/ED-2458)
[this bug](https://uktrade.atlassian.net/browse/ED-2460)


triage/templates/triage/wizard-step-base.html
```diff
@@ -11,7 +11,7 @@
     <section class="questions {% block container_class %}{% endblock %}">
         <div class="container">
             <div class="question ">
-				<form action="" method="post" enctype="multipart/form-data">
+				<form action="{% url 'triage-wizard' %}" method="post" enctype="multipart/form-data">
```
richtier: fun fact: before this change, the current url is used - so the browser could post to `/triage?result=true` - meaning when the user made their way back to the summary page having reviewed their answers...the summary would show their persisted answers, not their new answers.

I have added a test to check this no longer happens.

triage/templates/triage/wizard-step-summary.html
```diff
@@ -29,9 +33,5 @@ <h3 class="title">You told us:</h3>
 {% endblock %}
 
 {% block submit_button_label %}
-	{% if is_user_reviewing_persisted_answers %}
-		Continue my exporting journey
-	{% else %}
-		Create my exporting journey
-	{% endif %}
```
richtier: these carriage returns were causing odd UI: the button renders the carriage returns so the button is large.

PR #32, opened 2017-10-26
ED-2409 time left to read

sdonk:


article/helpers.py
```diff
@@ -8,26 +9,65 @@
...
 
+def filter_lines(lines_list):
+    """BeautifulSoup returns \n as lines as well, we filter them out.
+
+    It's a function because more filtering can be added later.
+    """
+    return filter(lambda x: x != '\n', lines_list)
+
+
+def lines_list_from_html(html):
```
konradko: wouldn't it be better to use .md files?

sdonk: Ideally yes, but I couldn't find a library able to filter out the language markup like `###`, `-` and the link syntax or just count the content words.
The quickest way was to convert it to HTML and use `BeautifulSoup` to digest it and eliminate the blank lines.

article/articles.py
```diff
@@ -2,14 +2,30 @@
...
+        self.title = title
+        self.keywords = keywords
+        self.tasks = tasks
+        self.markdown_file_path = markdown_file_path
+        self.url = url
+        self.parent = parent
+
+    @cached_property
+    def time_to_read(self):
+        return helpers.time_to_read_in_minutes(self)
```
konradko: why not just `time_to_read = helpers.time_to_read_in_minutes(self)` in `__init__`?

article/helpers.py
```diff
@@ -8,25 +9,93 @@
...
+    lines_list = lines_list_from_html(html)
+    filtered_lines_list = filter_lines(lines_list)
+    total_words_count = count_average_word_number_in_lines_list(
+        filtered_lines_list
+    )
+    return round(total_words_count / WORDS_PER_MINUTE)
+
+
+def total_time_to_read_multiple_articles(articles):
+    return sum((time_to_read_in_minutes(article) for article in articles))
```
konradko: `sum((article.time_to_read for article in articles))`

article/helpers.py
```diff
@@ -8,25 +9,93 @@
...
         self.request = request
 
     persist_article = abc.abstractproperty()
     retrieve_articles = abc.abstractproperty()
 
-    def article_read_count(self, group):
-        read_articles = frozenset(self.retrieve_articles())
-        articles_in_group = structure.ALL_GROUPS_ARTICLES_SETS[group]
+    @staticmethod
+    def articles_group(group_key):
```
konradko: not `article_group`?

article/helpers.py
```diff
@@ -8,25 +9,93 @@
...
+    @staticmethod
+    def article_from_key(article_key):
+        return structure.ALL_ARTICLES_DICT[article_key]
+
+    def articles_from_keys(self, articles_keys):
+        return (self.article_from_key(key) for key in articles_keys)
+
+    def read_articles_keys_in_group(self, group_key):
+        read_articles_uuids = frozenset(self.retrieve_articles())
+        articles_in_group = self.articles_group(group_key).articles_set
```
konradko: wouldn't be a bit easier without wrapping `structure.ALL_GROUPS_DICT[group_key]` in `articles_group`? Or maybe that helper method should be in `structure.py` as well

article/helpers.py
sdonk: Nice catch!

article/helpers.py
```diff
@@ -8,25 +9,93 @@
...
 
-    def article_read_count(self, group):
-        read_articles = frozenset(self.retrieve_articles())
-        articles_in_group = structure.ALL_GROUPS_ARTICLES_SETS[group]
+    @staticmethod
+    def articles_group(group_key):
+        return structure.ALL_GROUPS_DICT[group_key]
+
+    @staticmethod
+    def article_from_key(article_key):
```
konradko: maybe `article_from_uuid` would be more descriptive?

article/helpers.py
```diff
@@ -8,25 +9,93 @@
...
+        return structure.ALL_ARTICLES_DICT[article_key]
+
+    def articles_from_keys(self, articles_keys):
+        return (self.article_from_key(key) for key in articles_keys)
+
+    def read_articles_keys_in_group(self, group_key):
```
konradko: I had trouble with this before - what is `group_key`? Group name? Then maybe `group_name`

article/helpers.py
sdonk: I'll move the helper to `structure.py`, I like the wrapping in case the lookup changes in the future.

article/structure.py
```diff
@@ -1,10 +1,28 @@
...
+
+    def __init__(self, articles, key, title, url):
+        self.articles = articles
+        self.key = key
+        self.title = title
+        self.url = url
+
+    @cached_property
+    def articles_set(self):
+        return frozenset([article.uuid for article in self.articles])
```
konradko: ditto - could be in `__init__`

article/structure.py
```diff
@@ -1,10 +1,28 @@
...
+
+    @cached_property
+    def total_reading_time(self):
+        return sum(
```
konradko: ditto

article/structure.py
```diff
@@ -246,10 +264,8 @@
     ALL_ARTICLES,
 ]
 ALL_GROUPS_DICT = {group.key: group for group in ALL_GROUPS}
-ALL_GROUPS_ARTICLES_SETS = {
-    group.key: frozenset([article.uuid for article in group.articles]) for
-    group in ALL_GROUPS
-}
+ALL_ARTICLES_DICT = {article.uuid: article for
```
konradko: not a fan of using type names in vars. Would go with just `ALL_ARTICLES` or `ARTICLE_PER_UUID`, but it's not that big of a deal

article/articles.py
sdonk: Theoretically, the advantage of `cached_property` over the `__init__` is to avoid that the `time_to_read` value is accidentally set to something else. `cached_property` would create a property that cannot be changed. I think in this case is not very important. @konradko

article/helpers.py
sdonk: It's a group of articles, should it not be plural? I ask the help of @richtier

article/helpers.py
sdonk: The `ArticleGroup` property is called `key` though. I'd like to keep key here or change it to `name` in the `ArticleGroup` class as well

article/helpers.py
richtier: I would not have this method on the `BaseArticleReadManager` - it's just getting from a dictionary `structure.ALL_GROUPS_DICT`

but to answer the question - it gets a single `ArticleGroup` from `ALL_GROUPS_DICT`, so I would call it `get_article_group`.

sdonk: Actually, the same function already exists in `structure.py`. I'll delete this one!

article/articles.py
sdonk: Another reason to use `cached_property`: doing `time_to_read = helpers.time_to_read_in_minutes(self`) in `__init__` produces a circular import (thanks @richtier for debugging that). So I'll keep this as a property, it's cached anyway.

sdonk: Resolving the circular import is quite a big task and, for the reason I said above, I'd prefer to use `cached_property`. I created a dedicated ticket for the circular import https://uktrade.atlassian.net/browse/ED-2475

PR #33, opened 2017-10-26
Added missing links to case studies

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2470)


casestudy/casestudies.py
```diff
@@ -1,10 +1,12 @@
 from collections import namedtuple
-
+from django.urls import reverse_lazy
```
sdonk: nice!

PR #35, opened 2017-10-27
Add link to contact service

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-23570)

image


core/templates/core/base.html
```diff
@@ -53,6 +53,11 @@
         <main id="content" class="{% block css_layout_class %}{% endblock css_layout_class %}" role="main">
             {% block content %}{% endblock %}
         </main>
+        <section class="error-reporting">
+            <div class="container">
+                <a href="https://contact-us.export.great.gov.uk/export_opportunities/FeedbackForm" target="_blank">Is there anything wrong with this page?</a>
```
konradko: Doesn’t look correct - that’s a form for export opportunities. Should also be an env var

richtier: ah yes oops. should be `https://contact-us.export.great.gov.uk/export_readiness/FeedbackForm`

I'll make it an env var but default it to `https://contact-us.export.great.gov.uk/export_readiness/FeedbackForm`

PR #36, opened 2017-10-27
Add counter to articles and article groups

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2416)

image

## without '?source=...' in the url
image

## with '?source=...' in the url
image

### DIFF
--- a/article/views.py
+++ b/article/views.py
@@ -8,12 +8,12 @@ from core.helpers import build_social_links
 class ArticleReadCounterMixin:
     def get_article_read_counter(self):
         manager = helpers.ArticleReadManager(request=self.request)
-        uuids = set([item.uuid for item in self.article_group.articles])
-        read_uuids = set(manager.retrieve_articles())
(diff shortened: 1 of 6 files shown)


article/views.py
```diff
@@ -5,7 +5,25 @@
 from core.helpers import build_social_links
 
 
-class BaseArticleDetailView(TemplateView):
+class ArticleReadCounterMixin:
+    def get_article_read_counter(self):
+        manager = helpers.ArticleReadManager(request=self.request)
+        uuids = set([item.uuid for item in self.article_group.articles])
+        read_uuids = set(manager.retrieve_articles())
```
sdonk: This PR https://github.com/uktrade/directory-ui-export-readiness/pull/32 has exactly the same function in manager https://github.com/uktrade/directory-ui-export-readiness/pull/32/files#diff-e55fb4c2a9750982fa1302efd14288b0R64

If it gets merged first, then it can be used

PR #51, opened 2017-11-01
ED-2487 update header/footer

eadpearce:
Header/footer now updated with dev urls 
https://uktrade.atlassian.net/browse/ED-2487


makefile
```diff
@@ -117,8 +117,20 @@ DEBUG_SET_ENV_VARS := \
 	export COMPANIES_HOUSE_CLIENT_SECRET=debug-client-secret; \
 	export SECURE_HSTS_SECONDS=0; \
 	export PYTHONWARNINGS=all; \
-	export PYTHONDEBUG=true
-
+	export PYTHONDEBUG=true; \
+	export EXPORTING_NEW=/new; \
+	export EXPORTING_OCCASIONAL=/occasional; \
```
konradko: All these env vars also need adding to https://github.com/uktrade/directory-ui-export-readiness/pull/51/files#diff-cd5dbf3629c558b0104ba8f0937d6816L36 with a prefix `DIRECTORY_UI_EXPORT_READINESS_` - annoying, but needed for tests

ui/settings.py
```diff
@@ -216,6 +217,27 @@
 SECURE_HSTS_SECONDS = int(os.getenv('SECURE_HSTS_SECONDS', '16070400'))
 SECURE_HSTS_INCLUDE_SUBDOMAINS = True
 
+# HEADER/FOOTER URLS
+# EXPORTING PERSONAS
+EXPORTING_NEW = os.environ["EXPORTING_NEW"]
```
konradko: We should default to prod links on all these

eadpearce: the header/footer context processor does that if the link isn't specified. they're listed in directory-constants https://github.com/uktrade/directory-constants/blob/master/directory_constants/constants/urls.py

makefile
eadpearce: okay. will fix

ui/settings.py
eadpearce: okay this was also breaking the tests. have fixed now

ui/settings.py
```diff
@@ -11,6 +11,8 @@
 """
 
 import os
+# import all production urls from directory_constants
+from directory_constants.constants import urls
```
richtier: we can avoid the above comment by giving the variable a explanatory name:

```
from directory_constants.constants import urls as default_urls
```

PR #59, opened 2017-11-02
Design tweaks on landing page

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2495)

image

image


article/structure.py
```diff
@@ -23,7 +23,7 @@ def total_reading_time(self):
 
 
 PERSONA_NEW_ARTICLES = ArticleGroup(
-    key='persona-new',
+    key='persona_new',
```
konradko: Maybe just “name” instead of “key”? Bit confusing as this not a key-value struct

PR #61, opened 2017-11-02
Add privacy and about

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2505)

DNM - these pages might not be served from this app after all.

image

image


konradko:
Will reopen with conflicts fixed

PR #63, opened 2017-11-03
Custom page design tweaks + read manager refactor

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2506)
[this task](https://uktrade.atlassian.net/browse/ED-2519)
[this task](https://uktrade.atlassian.net/browse/ED-2530)

image


article/helpers.py
```diff
@@ -59,24 +59,22 @@ def __init__(self, request):
         self.request = request
 
     persist_article = abc.abstractproperty()
-    retrieve_articles = abc.abstractproperty()
+    retrieve_article_uuids = abc.abstractproperty()
```
konradko: not `abstractmethod`?

article/helpers.py
```diff
@@ -86,6 +84,7 @@ def remaining_reading_time_in_group(self, group_name):
         """ Return the remaining reading time in minutes
             for unread articles in the group
         """
+
         read_articles_uuids = self.read_articles_keys_in_group(group_name)
```
sdonk: It's not relevant to your code, but I'd change this to `read_articles_uuids_in_group` for consistency

core/tests/test_core_middleware.py
```diff
@@ -4,11 +4,11 @@
 
 from django.http import HttpResponse
 
-from ui import middleware
```
konradko: don't think we need `test_core_middleware.py` name, it's already under `core` folder

article/helpers.py
richtier: ah yes, it should be

core/tests/test_core_middleware.py
richtier: pytest sometimes gets confused when it encounters test files with the same name and something like "tried to import blah but __file__ is not what we expected"

richtier:
The failing code cov is code that will never run - inside an abstract method.

PR #68, opened 2017-11-06
Fix top importer value

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2540)

image


triage/tests/test_views.py
```diff
@@ -710,3 +710,24 @@ def test_custom_page_top_markets(sector_code, client):
         url = reverse('custom-page')
         response = client.get(url)
         assert response.status_code == 200
+
+        soup = BeautifulSoup(response.content, 'html.parser')
+        top_import_name = soup.find(id='top_importer_name')
+        
+        # data is not available for service codes, only for Harmonised
+        # System codes.
+        if not sector_code.startswith('HS'):
```
sdonk: Not a fan of `if not` and this doesn't save on indentation because the `else` branch is indented anyway

PR #93, opened 2017-11-13
Next section article link

eadpearce:
https://uktrade.atlassian.net/browse/ED-2598


article/templates/article/detail-base.html
```diff
@@ -37,6 +37,15 @@
             <span class="verbose">:&nbsp;</span><a id="next-article-link" href="{{ next_article.url }}?source={{ article_group.name }}">{{ next_article.title }}</a>
           </p>
         </div>
+    {% else %}
+      {% if article_group.next_guidance_group %}
+      <div class="link-continuity">
+        <p>
+          <span class="lead">Next</span>
+          <span class="verbose">:&nbsp;</span><a id="next-article-link" href="{{ article_group.next_guidance_group.articles.0.url }}?source={{ article_group.next_guidance_group.name }}">{{ article_group.next_guidance_group.articles.0.title }}</a>
```
sdonk: This should probably happen in the BE. What's the URL this is supposed to link to? The first article in the next group?

eadpearce: yep that's right

PR #100, opened 2017-11-15
ED2634 Add great urls and views

sdonk:


core/tests/test_views.py
```diff
@@ -12,7 +12,7 @@ def test_landing_page(client):
     response = client.get(url)
 
     assert response.status_code == 200
-    assert response.template_name == [views.LandingPagelView.template_name]
+    assert response.template_name == [views.LandingPageView.template_name]
```
konradko: nice :-)

PR #101, opened 2017-11-15
Add privacy and about pages

konradko:


core/tests/test_views.py
```diff
@@ -63,3 +63,21 @@ def test_robots(client):
...
+def test_about(client):
+    url = reverse('about')
+
+    response = client.get(url)
+
+    assert response.status_code == 200
+    assert response.template_name == ['core/about.html']
+
+
+def test_privacy(client):
```
sdonk: If are we going to write this test for all the great static pages it would be nice to parametrise it.

PR #111, opened 2017-11-17
Ed 2630 international landing page

stevenburnell-dit:
**Desktop**
screen shot 2017-11-17 at 17 37 06-fullpage

**Mobile**
screen shot 2017-11-17 at 17 36 53-fullpage


core/sass/components/_header.scss
```diff
@@ -0,0 +1,175 @@
+// Main Site Header
```
konradko: We are using https://github.com/uktrade/directory-header-footer for header/footer on FAB/FAS/Exred/SSO/SOO - @eadpearce can advise on that.

If this is for the international page, the naming of the files and descriptions need to reflect that.

stevenburnell-dit: The site is supposed to be using components/header and components/footer directly within the same repo. However, since directory-ui-export-readiness hasn't been built properly using the structure and files supplied from exred repo, I've now simply incorporated the specific CSS from those shared _header.scss and _footer.scss files directly into the international.scss

core/templates/core/base_international.html
```diff
@@ -0,0 +1,101 @@
...
+</head>
+<body id="top">
+    {% include 'core/google_tag_manager_body.html' %}
+    <header id="header">
+        <a id="skip-link" href="#content" class="verbose">Skip to main content</a>
+        <section role="dialog" data-component="language-selector-dialog" data-lang="en" data-label="International" data-name="English" class="language-selector-dialog">
+            <div data-node="languages" class="languages">
+                <p id="great-languages-selector">Choose your language</p>
+                <ul aria-labelledby="great-languages-selector" class="navigation">
+                <li><a href="?lang=en">English</a></li>
```
konradko: these should all lead to`/international/??lang=`

stevenburnell-dit: The currently specified href value means it would just add ?lang=xx param to current URL, but have updated code to reflect your preferred "/international/?lang=" specific location.

PR #124, opened 2017-11-22
Allow returning to custom page from articles

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2490)


article/tests/test_views.py
```diff
@@ -320,7 +319,8 @@ def test_articles_share_links(view_class, url, client):
 
 
 # skip the last group - it does not have a page, it's a list of all articles.
-@pytest.mark.parametrize('group', structure.ALL_GROUPS[:-1])
+# skip CUSTOM_PAGE_* groups - they're tested elsewhere as they need settting up
```
konradko: double comment it looks!

richtier: both are valid through - the last item is skipped, and the the three CUSTOM_PAGE_* groups are also skipped

PR #125, opened 2017-11-22
Ed 2530 persona page

stevenburnell-dit:
Design tweaks mainly affecting the spacing around Hero element.


richtier:
@stevenburnell-dit with design changes please include print screens in the PR

PR #129, opened 2017-11-23
Add etag header on static pages for CDN caching

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2635)

## triage cookie not set

image

## triage cookie set

image

todo:
 - [x] port conditional button text on landing page to javascript
 - [x] add etags to all static pages
 - [x] test


triage/views.py
```diff
@@ -169,7 +170,15 @@ def dispatch(self, request, *args, **kwargs):
                 kwargs={'step': TriageWizardFormView.SECTOR}
             )
             return redirect(url)
-        return super().dispatch(request, *args, **kwargs)
+        response = super().dispatch(request, *args, **kwargs)
+        response.set_cookie(
```
richtier: I set the cookie on the custom page view, rather than on triage complete view.

setting the cookie on triage completed would result in the cookie being invalidated too early: if the user has not been to the triage for a long time their cookie will be invalidated, but if they visit the custom page often then it wont be lost.

core/templates/core/landing-page.html
```diff
@@ -112,3 +110,13 @@ <h2 class="heading">Grow your business</h2>
         </div>
     </section>
 {% endblock %}
+
+{% block bodyScripts %}
+    <script type="text/javascript">
+      var hasCompletedTriage = document.cookie.indexOf('{{ TRIAGE_COMPLETED_COOKIE_NAME }}=true') > -1;
```
konradko: would be good to have a comment why we do that (cause page needs to be static)

core/views.py
```diff
@@ -26,15 +28,21 @@ def get_context_data(self, *args, **kwargs):
...
+class LandingPageView(SetEtagMixin, TemplateView):
     template_name = 'core/landing-page.html'
 
     def get_context_data(self, *args, **kwargs):
-        answer_manager = TriageAnswersManager(self.request)
-        has_completed_triage = answer_manager.retrieve_answers() != {}
         return super().get_context_data(
             *args, **kwargs,
-            has_completed_triage=has_completed_triage,
+            TRIAGE_COMPLETED_COOKIE_NAME=settings.TRIAGE_COMPLETED_COOKIE_NAME,
```
konradko: this will need to be added to privacy-cookie page - please let PM know

triage/views.py
konradko: Maybe we should name the cookie more specifically (like the sso_display_login), as it should only be used on the domestic landing page to check if the triage was completed. A name discouraging from usage anywhere else would be best, something like `DO_NOT_USE_THIS_ANYWHERE_ELSE_THAN_STATIC_LANDING_PAGE_COMPLETED_TRIAGE` ;-)

We also need a test checking that DB is not accessed during rendering of the static pages

ui/settings.py
```diff
@@ -268,6 +268,9 @@
 SESSION_COOKIE_SECURE = os.getenv('SESSION_COOKIE_SECURE', 'true') == 'true'
 SESSION_COOKIE_HTTPONLY = True
 CSRF_COOKIE_SECURE = True
+TRIAGE_COMPLETED_COOKIE_NAME = os.getenv(
```
konradko: do we expect this to be different depending on env? If not no need for env var

core/views.py
```diff
@@ -26,32 +28,37 @@ def get_context_data(self, *args, **kwargs):
...
             casestudies=[
                 casestudies.MARKETPLACE,
                 casestudies.HELLO_BABY,
                 casestudies.YORK,
             ],
-            article_group_read_progress=(
-                self.request.article_read_manager.get_group_read_progress()
-            ),
```
richtier: removed "3/5 completed" from the landing page

image

we can add a ticket to port this behaviour over to javascript.

PR #132, opened 2017-11-23
Ed 2710 international page

stevenburnell-dit:
Update placeholder images to those supplied by design
Also includes fixes (spacing and font) to reflect Zeplin design.


core/sass/partials/_effects.scss
```diff
@@ -132,18 +132,17 @@
...
+    cursor: pointer;
+    font-size: inherit;
     padding: 0;
     position: absolute;
-    right: $spacing-4;
-    top: $spacing-4;
-    width: 30px;
+    right: 0;
+    text-decoration: underline;
+    top: 0;
```
eadpearce: (can't review this specific change because it was in the previous PR) just noticed in the zeplin that the modal overlay is black with .4 opacity. does the style for .Modal-Overlay (line 157) want to be updated to match?

core/templates/core/base_international.html
```diff
@@ -52,15 +52,14 @@
 
         <div id="header-bar">
             <div class="container">
-                <p class="government-links"><img src="{% static 'images/international/df-it-186-digi-aw-white.svg' %}" alt="Department for International Trade" height="20">
-                </p>
+                <p class="government-links"></p>
```
eadpearce: (comment for line 36) would it make more sense for the p tag here to be an h2 instead? since it has a subheading on the domestic page version - so we could have an h2 and then a p tag there https://app.zeplin.io/project/58ac6a5b11c1d45f91a7d47c/screen/5a0461d2e0f018aafd15c2b5

core/sass/partials/_effects.scss
stevenburnell-dit: Have corrected overlay.

PR #133, opened 2017-11-23
Ed 2680 trans international pages

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2680)

## german

image

## english

image


locale/fr/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,151 @@
+# SOME DESCRIPTIVE TITLE.
```
konradko: no FR version

PR #161, opened 2017-11-28
Ed 2747 language select

eadpearce:
Updated directory-header-footer version again to fix language icon


core/templates/core/landing-page.html
```diff
@@ -24,13 +24,13 @@
...
     <div data-node="languages" class="languages">
       <ul aria-labelledby="great-languages-selector" class="navigation">
-        <li><a href="http://export.great.gov.uk/international?lang=en">English (International)</a></li>
-        <li><a href="http://export.great.gov.uk/international?lang=zh">中国</a></li>
-        <li><a href="http://export.great.gov.uk/international?lang=de">Deutsch</a></li>
-        <li><a href="http://export.great.gov.uk/international?lang=ja">日本語</a></li>
-        <li><a href="http://export.great.gov.uk/international?lang=es">Español</a></li>
-        <li><a href="http://export.great.gov.uk/international?lang=pt">Português</a></li>
-        <li><a href="http://export.great.gov.uk/international?lang=ar">عربى</a></li>
+        <li><a href="{% url 'landing-page-international' %}?lang=en">English (International)</a></li>
```
richtier: can you please replace with the for loop like in https://github.com/uktrade/directory-ui-export-readiness/blob/deef7ef2b328b5ad1095bd7302097fffd7b4e68f/core/templates/core/base_international.html#L39

that way this template is guaranteed to have the same languages as defined in `settings.py` `LANGUAGES`

eadpearce: ah. didn't see this. thanks

PR #181, opened 2017-11-30
Add legacy redirects

konradko:


ui/url_redirects.py
```diff
@@ -0,0 +1,331 @@
...
+from django.views.generic.base import RedirectView
+
+import core.views
+
+
+redirects = [
+    url(
+        r'^invest/$',
+        RedirectView.as_view(
+            url='https://invest.great.gov.uk',
```
richtier: should this work with HTTP and HTTPS? it seems odd to include the scheme.

PR #189, opened 2017-12-01
Add geolocation redirection

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2791)


core/static/js/geolocation.js
```diff
@@ -0,0 +1,20 @@
...
+
+dit.geolocationRedirection = (new function() {
+  this.init = function() {
+    $.getJSON("//freegeoip.net/json/").done(function(data) {
+      if (data.country_code !== 'GB' && data.country_code !== 'IE') {
+        var countryToLanguageMap = {
+          CN: 'zh-hans',
+          DE: 'de',
+          ES: 'es',
+          JP: 'ja',
```
konradko: we have Brazil-pt as well - if adding ES JP maybe that one as well?

PR #190, opened 2017-12-01
Ed 2767 fix english uk link focus

stevenburnell-dit:


core/templates/core/landing-page.html
```diff
@@ -8,7 +8,6 @@
   <script src="{% static 'js/dit.components.languageSelector.js' %}"></script>
   <script src="{% static 'js/dit.components.video.js' %}"></script>
   <script src="{% static 'js/home.js' %}"></script>
-  <script src="{% static 'Dit-Pattern-Styling/public/js/script.js' %}"></script>
```
eadpearce: taking this out will remove recent updates to the header js (i.e. keyboard navigation in header menu). you can check the exred js against the script in pattern library for changes (not ideal I know) https://github.com/uktrade/great-pattern-library/blob/master/shared-header-footer/script.js

stevenburnell-dit: Have added the script.js back and removed the components.menu.init() call (as discussed in Slack).

PR #194, opened 2017-12-04
ED-2807 transfer article count into session

sdonk:


article/helpers.py
```diff
@@ -159,7 +159,7 @@ def transfer_article_progress(self):
         """ Get the read articles from session and
             copies them to db if not there.
         """
-        articles_uuids_in_db = self.read_article_uuids
+        articles_uuids_in_db = self.retrieve_article_uuids()
```
richtier: please add a comment linking to the bug this fixes

I can see future developers optimising the app and breaking it :)

PR #200, opened 2017-12-04
Update Export Status Selector (Occasional) image

stevenburnell-dit:


richtier:
oops we somehow both did the task task: https://github.com/uktrade/directory-ui-export-readiness/pull/202

PR #205, opened 2017-12-05
Bulk create articles read for performance gains

richtier:
[part of this task](https://uktrade.atlassian.net/browse/ED-2822)


article/helpers.py
```diff
@@ -147,26 +155,17 @@ def persist_article(self, article_uuid):
...
+
     def retrieve_article_uuids(self):
         response = api_client.exportreadiness.retrieve_article_read(
             sso_session_id=self.request.sso_user.session_id
         )
         response.raise_for_status()
         uuids = [article['article_uuid'] for article in response.json()]
         return frozenset(uuids)
-
-    def transfer_article_progress(self):
```
richtier: I changed the way we handle this for performance gains. It's may be not as nice, but it solves our performance problem.

## before
 - get a list of articles that have been according to the database
 - determine the articles that have been read according to the session, but not according to the database
 - create "article read" objects in series.

## after
- if there are any "articles read" in the session then bulk create them in the database
- clear the articles from the session once they are in the database

With *before* there would be a minimum of one GET per request, and potentially as many as 45 POSTs if there are a lot of articles to synchronise from the session.

With *after* there will be a minimum of zero requests, and potentially one POST request to bulk create the articles read.

article/helpers.py
```diff
@@ -133,40 +146,40 @@ def retrieve_article_uuids(self):
...
-        self.transfer_article_progress()
+    article_uuids = frozenset()
 
     def persist_article(self, article_uuid):
         response = api_client.exportreadiness.create_article_read(
             article_uuid=article_uuid,
             sso_session_id=self.request.sso_user.session_id,
         )
-        assert response.ok, response.content
+        log_response(response)
```
richtier: API will return 400 if the article has already been seen, but that's fine given this facilitates:
 - not having to retrieve a list of already viewed articles
 - not creating duplicate article views

PR #218, opened 2017-12-07
ED-2869 beta banner

eadpearce:
https://uktrade.atlassian.net/browse/ED-2869


core/templates/core/base_international.html
```diff
@@ -60,6 +60,7 @@
                 <p class="government-links"></p>
             </div>
         </div>
+        {% include 'core/widgets/beta.html' %}
```
richtier: please make the "beta" label be a feature flag thing

so in the header something like
```
{% if header_footer_show_beta_label %}
    {% include 'core/widgets/beta.html' %}
{% endif %}
```

then in `directory_header_footer.context_processors.header_footer_context_processor` add

 ```
'header_footer_show_beta_label': getattr(settings, 'HEADER_FOOTER_SHOW_BETA_LABEL', False)
```

eadpearce: it's not in the header/footer. do we want this just as a feature flag in exred?

PR #233, opened 2017-12-11
Added keyboard trap to modals

eadpearce:
https://uktrade.atlassian.net/browse/ED-2865


core/static/js/dit.class.modal.js
```diff
@@ -71,28 +72,32 @@
...
   }
-  
+
   Modal.createContent = function() {
     var $content = $(document.createElement("div"));
     $content.addClass(CSS_CLASS_CONTENT);
     return $content;
   }
 
+  Modal.findFirstLink = function($container) {
```
stevenburnell-dit: This selector is quite clearly targeting something specific to the Language Selector, so doesn't belong here. If this function is required, there should be a generic selector, e.g. `$("a", this)` (where `this` is a reference to the current modal node). An overwriting function, with a selector that is specific to the Language Selector should go in the LanguageSelectorDialog (or whatever it is called) class.

Of course, if the whole thing is specific to a Language Selector dialog and not a generic modal (e.g. something that also drives the Video modal), then perhaps the whole findFirstLink custom handling should just be kept out of the Modal class and added directly to the inheriting LanguageSelectorDialog class.

core/static/js/dit.class.modal.js
```diff
@@ -132,14 +137,28 @@
       }
     });
 
+    self.$firstLink.on("keydown", function(e) {
```
stevenburnell-dit: Ok, I see what you're doing with the $firstLink.on and additions to $closeButton events. 
We should be doing this different though. 

In Modal.prototype.focus we're getting the first focusable element. We should move this...
`self.$content.find("video, a, button, input, select").eq(0)`

...to within the constructor, and assign the return value (jQuery node reference) to an instance property, like you've done in `this.$firstLink`. It should obviously be renamed to something like $firstFocusElement, because it is not always going to be a link.

We then apply this on[keydown] event hander to that element, much like you have done., but it should not happen inside a class function called `Modal.bindCloseEvents `- that's simply confusing the code organisation.

Perhaps we should have some `Modal.bindKeyboardFocusEvents`, or something like that.

core/static/js/dit.class.modal.js
```diff
@@ -132,14 +137,28 @@
...
+
     self.$closeButton.on("click keydown", function(e) {
       // Close on click or Enter
       if(e.which === 1 || e.which === 13) {
         Modal.deactivate.call(self);
         e.preventDefault();
+      } else if (e.which === 9 && !e.shiftKey) {
+      // After pressing tab when close button is selected
+      // loop back around to the first link
+        self.$firstLink.focus();
```
stevenburnell-dit: This should just call self.focus(), since that should refocus to the required first element.

core/static/js/dit.class.modal.js
```diff
@@ -223,7 +242,6 @@
     var self = this;
     self.$content.find("video, a, button, input, select").eq(0).focus();
```
stevenburnell-dit: Taking into account the above comment, the `Modal.prototype.focus` function should be updated to simply be:
`self.$firstFocusElement.focus();`

core/static/js/dit.class.modal.js
```diff
@@ -154,7 +151,7 @@
       } else if (e.which === 9 && !e.shiftKey) {
       // After pressing tab when close button is selected
       // loop back around to the first link
-        self.$firstLink.focus();
+        self.focus();
```
stevenburnell-dit: Since you're adding the loop back code directly to self.$closeButton, doesn't that mean there's no point in finding and declaring a self.$lastFocusElement?

I might be missing it, but it doesn't look like $lastFocusElement is being used anywhere.

core/static/js/dit.class.modal.js
```diff
@@ -166,6 +163,18 @@
     }
   }
 
+  Modal.bindKeyboardFocusEvents = function() {
```
stevenburnell-dit: Perhaps it would be better to have a separate self.$lastFocusElement.on("keydown") specified/applied in this function to keep that trapping functionality together. It would also make things more readable and easier to understand the interaction between first and last.

core/static/js/dit.class.modal.js
```diff
@@ -237,10 +246,15 @@
     self.$content.append(content);
   }
 
-  // Tries to add focus to the first found element allowed nwith atural focus ability.
+  // Tries to add focus to the first found element allowed with natural focus ability.
   Modal.prototype.focus = function() {
     var self = this;
-    self.$content.find("video, a, button, input, select").eq(0).focus();
+    // If $firstFocusElement didn't find anything
```
stevenburnell-dit: Understand that you're being defensive with this check, and also the comment about appending elements, but would argue it is not handled in the best way. 

You're right about re-finding the first focus element after appending, but we should being doing something like:

1. Run `this.$firstFocusElement = Modal.findFirstFocusElement($container);` as another action of `Modal.prototype.setContent()` function
2. Add some kind of `this.$firstFocusElement.off()` (and for lastFocusElement) as opening (reset) lines for `Modal.bindKeyboardFocusEvents()` function 
3. Call `Modal.bindKeyboardFocusEvents()` as a last action to the `Modal.prototype.setContent()` function

That would clear things up and leave this function as simple and `self.$firstFocusElement.focus()`  
By all means leave the `!self.$firstFocusElement.length` check in if you want but I think, if `Modal.findFirstFocusElement` returns an empty jQuery object (so length ==0), then jQuery also handles a `.focus()` chained call without error as well, meaning the check isn't actually required. 

You can test that last assumption to be sure but, if you run this...
`$().focus()`
...in your browser console, you should see no JS error. That should be the same thing.

PR #239, opened 2017-12-12
Fix link to about-cookies

konradko:


core/tests/test_views.py
```diff
@@ -202,6 +202,16 @@ def test_privacy_view_domestic(client):
     ]
 
 
+def test_privacy_view_domestic_about_cookies_link_correct(client):
```
konradko: dumb test, but we'll add a smoke link-checker one

PR #240, opened 2017-12-12
Add security headers

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2913)


core/templates/core/base.html
```diff
@@ -7,6 +7,7 @@
         <meta id="utmCookieDomain" value="{{ analytics.UTM_COOKIE_DOMAIN }}" />
         {% include 'core/google_tag_manager_head.html' %}
         <meta charset="utf-8" />
+        <meta name="referrer" content="no-referrer">
```
konradko: why are we doing this?

richtier: django does not provide an out of the box way to add the "Referrer-Policy" header.

however, browsers support setting it via this `meta` approach.

So this is a valid but simpler way to achieve the same outcome.

PR #247, opened 2017-12-14
ED-2668 social share metadata

eadpearce:
https://uktrade.atlassian.net/browse/ED-2668


core/templatetags/image_tags.py
```diff
@@ -8,3 +9,14 @@
 def progressive_image(webp, src, alt_text=''):
     context = {'src': src, 'webp': webp, 'alt_text': alt_text}
     return render_to_string('core/widgets/progressive_image.html', context)
+
+
+class FullStaticNode(static.StaticNode):
```
konradko: why do we need this?

core/templatetags/image_tags.py
```diff
@@ -8,3 +9,14 @@
...
+    def url(self, context):
+        request = context['request']
+        return request.build_absolute_uri(super().url(context))
+
+
+@register.tag
+def get_image_absolute_uri(parser, token):
```
konradko: why not just use `%static%`?

eadpearce: opengraph needs the full path to the image. currently we're using `<meta property="og:image" content="{% static 'images/icons/opengraph-image.png' %}">` which produces `<meta property="og:image" content="/static/images/icons/opengraph-image.f86f1d0dd106.png">`

article/templates/article/detail-base.html
```diff
@@ -11,6 +12,23 @@
...
+  <meta name="description" content="" />
+
+  {# Twitter #}
+  <meta name="twitter:card" content="summary">
+  <meta name="twitter:title" content="{{ article.title }}">
+  {# twitter falls back to og for other data not included here #}
+
+  {# Facebook share data #}
+  <meta property="og:locale" content="en_UK" />
+  <meta property="og:site_name" content="great.gov.uk" />
```
richtier: will hard-coding this cause any problems if we try to share staging or dev urls (for testing purposes) ?

core/templatetags/image_tags.py
richtier: ah yes this is an annoying thing to solve with django:
- if we serve images from the webserver (local dev) then `{% static ... %}` gives a relative url.
- if we use AWS as the storage backend (prod, staging, heroku dev) then `{% static ... %}` gives an absolute url pointing to the s3 bucket (as defined by `AWS_S3_CUSTOM_DOMAIN` in `settings.py`.

Unfortunately `request.build_absolute_uri` will create a absolute url for the current domain, not the s3 domain that images are served from if s3 storage backend is used.

core/templates/core/base.html
```diff
@@ -7,17 +7,19 @@
         <meta id="utmCookieDomain" value="{{ analytics.UTM_COOKIE_DOMAIN }}" />
         {% include 'core/google_tag_manager_head.html' %}
         <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+
         <title>{% block head_title %}Welcome to great.gov.uk{% endblock %}</title>
 
+        {% block sharing_metadata %}{% endblock %}
```
richtier: we can define default value in here, so we can do:

```
{% block sharing_metadata %}
    <meta property="og:image" content="{% static 'images/icons/opengraph-image.png' %}">
{% endblock %}
```

which ensures an image will be specified for all pages

article/templates/article/detail-base.html
richtier: as discussed elsewhere: no :)

core/templates/core/base.html
eadpearce: aha yes i forgot about that

PR #251, opened 2017-12-18
Hide country column label from large screens but not screen readers

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2946)

we hide the country label as it's not needed for people who can see

they can see the column is for country

but for screen readers, we use the column label

## after fix
image

## before fix
image


ui/settings.py
```diff
@@ -249,7 +249,7 @@
 
 ANALYTICS_ID = os.getenv("ANALYTICS_ID")
 
-SECURE_SSL_REDIRECT = os.getenv('SECURE_SSL_REDIRECT', 'true') == 'true'
+SECURE_SSL_REDIRECT = False
```
konradko: Should not be changed

richtier: oy vey oops

PR #259, opened 2017-12-21
ED-2895 scope indicator

eadpearce:
https://uktrade.atlassian.net/browse/ED-2895


casestudy/templates/casestudy/base.html
```diff
@@ -42,7 +42,7 @@ <h1>{{ casestudy.title }}</h1>
                 {% block quote_two_link %}{% endblock %}
             </cite>
         </blockquote>
-        {% include 'core/social-share-links.html' %}
+        {% include 'core/widgets/social-share-links.html' %}
```
richtier: note the "widgets" folder initially started as a place to put templates for form field widgets.

But I see it's now the widgets folder has also got components to include in other template.

not a problem, just an FYI.

PR #263, opened 2017-12-24
Refactor article read manager

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2947)

- "read" can refer to _the action of reading_ or _the state of having been read_ - so it's confusing ot call something a `ReadManager` and methods that use the word `read`. To solve this instead use the words"view" and "viewed"
- the article read manager was too stateful. now it is less stateful.
- the "time to read article" code assumes each row has 5 words. No need to make that assumption.

### DIFF
--- a/article/helpers.py
+++ b/article/helpers.py
@@ -7,128 +7,92 @@ from django.template.loader import render_to_string
 from django.utils.functional import cached_property
 
 from api_client import api_client
-from . import structure
+from article import structure
 
 
 WORDS_PER_SECOND = 1.5  # Average word per second on screen
 
 
 def markdown_to_html(markdown_file_path, context={}):
-    html = render_to_string(markdown_file_path, context)
-    return markdown2.markdown(html, extras=["target-blank-links"])
+    content = render_to_string(markdown_file_path, context)
+    return markdown2.markdown(content, extras=["target-blank-links"])
 
 
-def filter_lines(lines_list):
-    """BeautifulSoup returns \n as lines as well, we filter them out.
-
-    It's a function because more filtering can be added later.
-    """
-    return filter(lambda x: x != '\n', lines_list)
-
-
-def lines_list_from_html(html):
-    """Parses the HTML to return text lines."""
-    soup = BeautifulSoup(html, 'html.parser')
-    return soup.findAll(text=True)
-
-
-def count_average_word_number_in_lines_list(lines_list, word_length=5):
-    """Assume average word length, counts how many words in all the lines."""
-    total_words = 0
-    for line in lines_list:
-        total_words += len(line)/word_length
-    return total_words
+def get_word_count(html):
+    soup = BeautifulSoup(html.replace('\n', ''), 'html.parser')
+    words = ''.join(soup.findAll(text=True)).strip()
+    return len(words.split(' '))
 
 
 def time_to_read_in_seconds(article):
-    """Return time to read in minutes give an Article object."""
     html = markdown_to_html(article.markdown_file_path)
-    lines_list = lines_list_from_html(html)
-    filtered_lines_list = filter_lines(lines_list)
-    total_words_count = count_average_word_number_in_lines_list(
-        filtered_lines_list
-    )
-    return round(total_words_count / WORDS_PER_SECOND)
+    return round(get_word_count(html) / WORDS_PER_SECOND)
 
 
-def total_time_to_read_multiple_articles(articles):
-    return sum((article.time_to_read for article in articles))
-
-
-class BaseArticleReadManager(abc.ABC):
-    def __init__(self, request, current_article=None):
+class BaseArticlesViewedManager(abc.ABC):
+    def __init__(self, request):
         self.request = request
-        self.current_article = current_article
 
-    @abc.abstractmethod
-    def retrieve_historic_article_uuids(self):
-        return
-
-    def get_group_read_progress(self):
+    def get_view_progress_for_groups(self):
         return {
             group.name: {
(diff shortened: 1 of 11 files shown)


richtier:
codecov is failing because an abstractmethod is not called.

article/articles.py
```diff
@@ -21,10 +21,10 @@ def __init__(self, uuid, title, keywords, tasks,
         self.parent = parent
 
     @cached_property
-    def time_to_read(self):
+    def time_to_view(self):
```
konradko: don't love the name, but understand the confusion with 'read' past tense :-)

article/helpers.py
```diff
@@ -7,128 +7,81 @@
...
+    def __init__(self, request):
         self.request = request
-        self.current_article = current_article
-
-    @abc.abstractmethod
-    def retrieve_historic_article_uuids(self):
-        return
 
-    def get_group_read_progress(self):
+    def view_progress_for_groups(self):
```
konradko: `get_view_progress_for_groups` would be nicer, as this ain't a property

article/helpers.py
```diff
@@ -7,128 +7,81 @@
...
-        # read_articles_in_category is a new set (intersection)
-        # with elements common to read_articles and articles_in_category
-        read_articles_in_group = self.read_article_uuids & article_uuids
-        return read_articles_in_group
-
-    def article_read_count(self, group_name):
-        return len(self.read_articles_keys_in_group(group_name))
+    def articles_viewed_for_group(self, group_name):
+        group = structure.get_article_group(group_name)
+        return self.viewed_article_uuids & group.article_uuids
```
konradko: please leave the comment about `&` returning intersection

article/helpers.py
```diff
@@ -7,128 +7,81 @@
...
+    def viewed_article_uuids(self):
+        return self.retrieve_viewed_article_uuids()
+
+    @abc.abstractmethod
+    def retrieve_viewed_article_uuids(self):
+        return
 
 
-class ArticleReadManager:
+class ArticlesViewedManager:
```
konradko: let's rename the model in directory-api to `ArticleViewed` as well

konradko: in terms of patterns, this is rather `ArticlesViewedManagerFactory`

article/helpers.py
```diff
@@ -7,128 +7,81 @@
...
     def __new__(cls, request, current_article=None):
-        session_manager = SessionArticlesReadManager(request, current_article)
-        database_manager = DatabaseArticlesReadManager(
-            request, current_article
-        )
+        session_manager = SessionArticlesViewedManager(request)
```
konradko: it is a bit of unexpected that instantiating a class can cause DB reads and writes. But I understand this is to handle the complexity of syncing session and db progress.

To keep it explicit let's split the code below into 2 paths - user logged in and not (e.g.

```
if request.sso_user is None:
    articles_viewed_manager = self.get_session_articles_viewed_manager(request, current_article)
else:
    articles_viewed_manager = self.get_database_articles_viewed_manager(request, current_article)

return articles_viewed_manager
```

article/helpers.py
```diff
@@ -7,128 +7,81 @@
...
+
         if request.sso_user is None:
             if current_article:
                 session_manager.persist_article(current_article.uuid)
             return session_manager
-        response = database_manager.bulk_persist_article(
-            article_uuids=session_manager.read_article_uuids
-        )
+
+        viewed_article_uuids = set()
```
konradko: would be good to comment on what this achieves or move it to a function with an explicit name (e.g. `sync_session_and_database_viewed_articles`)

article/articles.py
konradko: as discussed, let's keep it as `time_to_read` while changing ArticleRead to ArticleViewed

article/structure.py
```diff
@@ -12,16 +12,22 @@ def __init__(self, articles, name, title, url, next_guidance_group=None):
...
         self.url = url
         self.next_guidance_group = next_guidance_group
-        self.articles_set = frozenset(
+        self.article_uuids = frozenset(
             [article.uuid for article in self.articles]
         )
 
     @cached_property
-    def total_reading_time(self):
+    def view_time(self):
```
konradko: let's use consistent `read_time` and `view_progress`

article/tests/test_helpers.py
```diff
@@ -40,8 +40,8 @@ def test_database_create_article_read_calls_api(
 ):
     mock_bulk_create_article_read.return_value = create_response(200)
 
-    manager = helpers.DatabaseArticlesReadManager(sso_request)
-    manager.bulk_persist_article(article_uuids=['123'])
+    manager = helpers.DatabaseArticlesViewedManager(sso_request)
+    manager.persist_articles(article_uuids=['123'])
```
konradko: `save_article_views` or `persist_article_views` would be a better name, as it's the views, not the articles being persisted here.

article/tests/test_helpers.py
```diff
@@ -197,36 +196,36 @@ def test_article_read_manager_handles_no_read_articles_on_synchronisation(
...
-        helpers.ArticleReadManager(sso_request)
+        helpers.ArticlesViewedManagerFactory(sso_request)
 
     assert sso_request.session[session_key] == [1, 2, 3]
     assert str(execinfo.value) == 'validation error'
 
 
 @patch('api_client.api_client.exportreadiness.bulk_create_article_read')
-def test_database_remaining_reading_time_in_group(
+def test_database_remaining_view_time_for_group(
```
konradko: `view_time` > `read_time`

PR #272, opened 2018-01-10
ED-3100 Removed 'show more' button on article lists

eadpearce:
https://uktrade.atlassian.net/browse/ED-3100


richtier:
@eadpearce do we want to remove it from https://github.com/uktrade/directory-ui-export-readiness/blob/9bbe66f29d8f422fa8eae219946cc24d016b15ca/triage/templates/triage/custom-page.html#L202 too?

if so, we can also delete https://github.com/uktrade/directory-ui-export-readiness/blob/master/article/static/js/article-list.js

eadpearce:
@richtier yes! I thought I'd forgotten one other place where it was used

PR #273, opened 2018-01-11
ED-2958 Encode social links on international page

eadpearce:
https://uktrade.atlassian.net/browse/ED-2958


core/templates/core/base_international.html
```diff
@@ -1,5 +1,6 @@
 {% load static from staticfiles %}
 {% load trans from i18n %}
+{% load social_tags %}
```
richtier: to make it easier to determine what is being imported here please do something like:

```py
{% from social_tags import build_twitter_link etc %}
```

core/templatetags/social_tags.py
```diff
@@ -0,0 +1,32 @@
...
+import urllib.parse
+
+register = template.Library()
+
+
+@register.simple_tag
+def build_twitter_link(text, url):
+    encoded_text = urllib.parse.quote(text + ' ')
+    link = 'https://twitter.com/intent/tweet?text={}{}'
+    return link.format(encoded_text, url)
```
richtier: please add tests for these tags. examples [here](https://github.com/uktrade/directory-ui-supplier/blob/master/company/tests/test_templatetags.py)

core/templates/core/base_international.html
```diff
@@ -1,5 +1,9 @@
 {% load static from staticfiles %}
 {% load trans from i18n %}
+{% load build_email_link from social_tags %}
+{% load build_linkedin_link from social_tags %}
+{% load build_twitter_link from social_tags %}
+{% load build_facebook_link from social_tags %}
```
richtier: this is ok. note in future you caan do `{% load build_email_link build_linkedin_link from social_tags %}`

core/templatetags/tests/test_social_tags.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_build_social_links():
+    actual = social_tags.build_twitter_link(
+        'Welcome to the UK',
+        'http://exred.trade.great')
+    expected = (
+        'https://twitter.com/intent/tweet?text=Welcome%20to%20the%20UK%20'
+        'http://exred.trade.great')
+    assert(actual == expected)
```
richtier: assert is not a function - so for consistent style can you please change this to:

```
assert actual == expected
```

core/templates/core/base_international.html
eadpearce: ah! didn't know that

PR #277, opened 2018-01-15
Add instructional page to triage

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3005)

image


triage/templates/triage/start-now.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block content %}
+	<div class="container">
+		<div  id="start-now-container">
+			<h1>Create your export journey</h1>
+			<br>
+			<p>
+				Answer a few quick questions about what you want to export and your level of experience and get:
+				<ul>
+					<li>guidance aimed at your experience level</li>
```
eadpearce: can we use spaces instead of tabs here? all the other templates seem to use spaces

PR #294, opened 2018-01-25
ED-3205 custom page redirect

eadpearce:
https://uktrade.atlassian.net/browse/ED-3205


ui/settings.py
```diff
@@ -264,6 +264,8 @@
     'GREAT_EXPORT_HOME', default_urls.GREAT_EXPORT_HOME)
 
 # EXPORTING PERSONAS
+CUSTOM_PAGE = os.getenv(
+    'CUSTOM_PAGE', default_urls.CUSTOM_PAGE)
```
richtier: we no longer need to default to `default_urls.*` - its handled by the context processor in header footer.

could you please update the settings.py so we do
```
os.getenv('CUSTOM_PAGE')
```

for here, and the existing usages of `default_urls` in the file.

eadpearce: oh yeah. i just copypasted. i can update the other settings as well

PR #308, opened 2018-02-09
ED-3377 fix missing style

eadpearce:
https://uktrade.atlassian.net/browse/ED-3377


core/templates/core/not_found.html
```diff
@@ -10,11 +10,11 @@
 {% block content %}
```
richtier: is this file used? I think the `404.html` is being used. If `not_found.hyml` is not being used feel free to delete this.

eadpearce: nope. i thought 404 was being used but actually it's this one. the 404 template doesn't render correctly so i'm leaving it like this. i can raise this bug in a separate ticket

PR #319, opened 2018-02-27
ED-3593 Fix some links disappearing when focused

eadpearce:
https://uktrade.atlassian.net/browse/ED-3593


core/sass/components/_carousel.scss
```diff
@@ -3,6 +3,9 @@
 
 .carousel-cta {
     @include continue-arrow;
+    &:focus {
```
richtier: do we also need `:visited`?

eadpearce: not for the arrows i don't think. the visited style for the arrows is a purple tick mark which is used for articles when they've been read. i don't think that's being used for case studies

PR #325, opened 2018-03-06
ED-3647 Remove sector question from triage, hide table on custom page

eadpearce:
https://uktrade.atlassian.net/browse/ED-3647


triage/tests/test_views.py
```diff
@@ -26,14 +26,9 @@ def mock_retrive_articles_read():
 @patch('triage.helpers.SessionTriageAnswersManager.persist_answers')
 def test_submit_triage_regular_exporter(mock_persist_answers, client):
     view_class = views.TriageWizardFormView
-    url = reverse('triage-wizard', kwargs={'step': view_class.SECTOR})
```
konradko: I believe the sector has to be removed from the serializer as well - in API, check api-client too.

eadpearce: we're going to be using it in future though just not in triage. does it still need removing?

konradko: probably not, but should be made optional - in the model as well

eadpearce: yeah it's already optional in the model and serializer

PR #326, opened 2018-03-08
ED-3612 Add goods/services question to triage

eadpearce:
Styling is wonky at the moment. Will be fixed in this ticket: https://uktrade.atlassian.net/browse/ED-3613
https://uktrade.atlassian.net/browse/ED-3612


kowalcj0:
LGTM, I'd appreciate if you could add IDs to various page elements.
whatdoyouwanttoexport

PS. here's the ticket for adding IDs on other Triage pages https://uktrade.atlassian.net/browse/ED-3681

triage/forms.py
```diff
@@ -126,11 +148,21 @@ def get_used_marketplace(answers):
     return answers.get('used_online_marketplace') is True
 
 
+def get_exporting_goods(answers):
```
konradko: is this needed? Does not seem to be used anywhere

eadpearce: that is true. i thought it'd be needed but isn't currently. will remove

PR #338, opened 2018-04-09
ED-4015 Custom page table

eadpearce:
https://uktrade.atlassian.net/browse/ED-4015

screencapture-exred-trade-great-8007-custom-2018-04-10-15_05_49


triage/forms.py
```diff
@@ -136,6 +145,26 @@ class CompaniesHouseSearchForm(forms.Form):
     term = forms.CharField()
 
 
+class SectorForm(forms.Form):
+    sector = forms.ChoiceField(
```
richtier: you can now use export-components's `fields.ChoiceField` - no need to set `label_suffix` and adds some style classes for us

triage/forms.py
```diff
@@ -136,6 +145,26 @@ class CompaniesHouseSearchForm(forms.Form):
...
+        choices=BLANK_CHOICE_DASH + SECTORS_CHOICES,
+        label='',
+        label_suffix='',
+        widget=forms.Select(attrs={'id': 'js-sector-select'}),
+    )
+
+    def update_triage_sector(self, request, answers):
+        answers.update(self.cleaned_data)
+        answer_manager = TriageAnswersManager(request)
+        answer_manager.persist_answers(answers)
```
richtier: it's a bit funky performing IO (in this case, saving to the database) inside the form.

I would move the view, where it's more obvious to the reader a side-effect is happening.

triage/forms.py
```diff
@@ -136,6 +145,26 @@ class CompaniesHouseSearchForm(forms.Form):
...
+
+
+def get_sector_label(cleaned_data):
+    for key, label in exred_sector_names.SECTORS_CHOICES:
```
richtier: what will happen if `cleaned_data` contains a value that is not in `exred_sector_names.SECTORS_CHOICES`.

passing in `cleaned_data` but accessing only `sector` is a but like asking for a banana but getting a gorilla holding a banana (I love that analagy). Can you pass in `sector` instead of `cleaned_data`?

triage/templates/triage/custom-page.html
```diff
@@ -40,6 +41,102 @@ <h2 class="heading-large">Connect with international buyers</h2>
...
+            {% if triage_result.sector and not service_sector %}
+            <p>Select a different product category.</p>
+            {% else %}
+            <p>Type the product you want to export to get statistics on the largest importers.</p>
+            {% endif %}
+            <div class="form-group{% if service_sector %} form-group-error{% endif %}">
+              {% if service_sector %}
+              <p>No data available, please select a different commodity.</p>
+              {% endif %}
+                {% csrf_token %}
```
richtier: just a note on convention: csrf_token is normally the first thing just after `<form>`, so 

```
<form>
{% csrf_token %}
...
```

as it's a convention lazy devs reading it may thing `csrf_token` is not present as it's not the first thing after `<form>`

triage/templates/triage/custom-page.html
```diff
@@ -40,6 +41,102 @@ <h2 class="heading-large">Connect with international buyers</h2>
...
+    </div>
+  </div>
+</section>
+{% endif %}
+{% if triage_result.is_exporting_goods %}
+<section id="sector">
+  <div class="container">
+    <div class="grid-row">
+      <div class="column-half">
+        <form action="/custom/" method="post">
```
richtier: if this template is only used on `/custom/` then this can be blank.

if the template is used in other places too it's better to use` {% url '<the-url-name' %}`, which you can get from urls..py

triage/views.py
```diff
@@ -162,24 +170,37 @@ class TriageStartPageView(TemplateView):
...
         return super().dispatch(request, *args, **kwargs)
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['persona'] = forms.get_persona(self.triage_answers)
         context['triage_result'] = self.triage_answers
+        # for people who selected a service category before the sector
+        # question was removed from triage
+        if self.triage_answers[
+                'sector'] and self.triage_answers['sector'].startswith('EB'):
```
richtier: bit hard to read this. consider something like 

```
context['service_sector'] = self.triage_answers.get('sector', '').startswith('EB')
```

triage/views.py
```diff
@@ -190,6 +211,14 @@ def get_context_data(self, **kwargs):
         context['article_group_read_progress'] = (
             self.article_read_manager.get_view_progress_for_groups()
         )
+        if self.triage_answers['sector'] is not None:
```
richtier: if `sector` is somehow an empty string then the code below will not be hit. would it be valid to instead check if `sector` is falsey with `if self.triage_answers['sector']:` ?

eadpearce: no because then you get a key error. i could check for both

triage/views.py
eadpearce: that would be nicer but it gives me `'NoneType' object has no attribute 'startswith'`

PR #340, opened 2018-04-11
ED-3619 T&C CMS

eadpearce:
https://uktrade.atlassian.net/browse/ED-3619


ui/settings.py
```diff
@@ -343,3 +343,8 @@
         'promo-video_web.mp4'
     )
 )
+
+# directory CMS
+CMS_URL = os.environ['CMS_URL']
+CMS_SIGNATURE_SECRET = os.environ['CMS_SIGNATURE_SECRET']
+FEATURE_CMS_ENABLED = os.getenv('FEATURE_CMS_ENABLED', 'true') == 'true'
```
richtier: we want this to be off by default, so `FEATURE_CMS_ENABLED = os.getenv('FEATURE_CMS_ENABLED', 'false') == 'true'`

ui/settings.py
```diff
@@ -343,3 +343,8 @@
         'promo-video_web.mp4'
     )
 )
+
+# directory CMS
+CMS_URL = os.environ['CMS_URL']
+CMS_SIGNATURE_SECRET = os.environ['CMS_SIGNATURE_SECRET']
+FEATURE_CMS_ENABLED = True
```
richtier: this should be feature flagged and off by default.

eadpearce: whoops!

PR #346, opened 2018-04-23
Add privacy-and-cookies CMS page, remove non-CMS terms-and-conditions…

konradko:
… page


ui/urls.py
```diff
@@ -49,22 +49,22 @@
...
     url(
         r"^privacy-and-cookies/$",
-        core.views.PrivacyCookiesDomestic.as_view(),
+        core.views.PrivacyCookiesDomesticCMS.as_view(),
         name='privacy-and-cookies'
     ),
     url(
         r"^terms-and-conditions/$",
-        core.views.TermsConditionDomesticViewNegotiator.as_view(),
+        core.views.TermsConditionsDomesticCMS.as_view(),
```
richtier: no feature flag?

core/tests/test_views.py
```diff
@@ -154,27 +155,26 @@ def test_terms_conditions_cms_disabled(
...
+            'privacy-and-cookies-international',
+            'core/privacy_cookies-international-cms.html'
         ),
     )
 )
-@patch('core.views.cms_client.export_readiness.get_terms_and_conditions_page')
-def test_terms_conditions_cms_enabled(
-    mock_get_t_and_c_page, view, expected_template, client, settings
+@patch('core.views.cms_client.export_readiness.get_privacy_and_cookies_page')
+def test_privacy_cookies_cms(
```
richtier: it looks like this has replaced the terms and conditions test?

konradko: t&c is still up `test_terms_conditions_cms`

ui/urls.py
konradko: no need for one - with the first page added it was important to be able to roll back quickly, but now we can just keep replacing.

PR #356, opened 2018-05-23
Add cms get finance page and styling

eadpearce:
depends on this PR for cms page https://github.com/uktrade/directory-cms/pull/91

https://uktrade.atlassian.net/browse/CMS-81

screencapture-exred-trade-great-8007-get-finance-2018-05-23-10_33_32


core/views.py
```diff
@@ -232,3 +234,18 @@ def get_context_data(self, *args, **kwargs):
...
 class TermsConditionsInternationalCMS(TermsConditionsDomesticCMS):
     template_name = 'core/terms_conditions-international-cms.html'
+
+
+class GetFinanceCMS(TemplateView):
+    template_name = 'core/get_finance.html'
+
+    def get_context_data(self, *args, **kwargs):
+        response = cms_client.lookup_by_slug(
+            slug='get-finance',
```
richtier: small layering problem: here exred "knows" the slug of the CMS page.

instead, we can put the slug [here](https://github.com/uktrade/directory-cms-client/blob/master/directory_cms_client/constants.py), and then the code would be:

```
cms_client.lookup_by_slug(
    slug=EXPORT_READINESS_GET_FINANCE_SLUG
```

the motivation is so the reader can infer a few things:
 - it's a constant (so should not change)
 - it "belongs" to the CMS, and CMS decides what the value should be.

Also, if ever we change a slug, no code change is needed in ExRed - just in `cms-client` (and in requirements.txt)

eadpearce: made PR for that so will update this once that's merged https://github.com/uktrade/directory-cms-client/pull/15

PR #372, opened 2018-06-14
Fix typos

eadpearce:


sdonk:
Pytest cache folder and files shouldn't be here

PR #377, opened 2018-06-22
Add guidance notes page, refactor

eadpearce:
https://uktrade.atlassian.net/browse/CMS-135


core/views.py
```diff
@@ -205,81 +206,36 @@ class AboutView(SetEtagMixin, TemplateView):
...
-            *args, **kwargs
-        )
+class PrivacyCookiesDomesticCMS(GetCMSPageMixin, TemplateView):
+    template_name = 'core/info_page.html'
+    slug = EXPORT_READINESS_PRIVACY_AND_COOKIES_SLUG
 
 
 class PrivacyCookiesInternationalCMS(PrivacyCookiesDomesticCMS):
-    template_name = 'core/privacy_cookies-international-cms.html'
+    template_name = 'core/info_page_international.html'
```
konradko: is the name of the page changing? Maybe renaming the view would make sense then, it's nice when template filenames are similar to the views that use them.

eadpearce: name of the page isn't changing so the view name can stay the same.`info_page_international` template is used by both the T&Cs and cookies pages. several pages now use a similar template (cookies, T&Cs, dashboard notes) so this is just to maximise code reuse

konradko: ah, nice!

PR #380, opened 2018-06-27
Move geolocation logic to backend

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-5)


core/helpers.py
```diff
@@ -58,3 +62,56 @@ def handle_cms_response(response):
...
+    def should_redirect(self):
+        return (
+            self.COOKIE_NAME not in self.request.COOKIES and
+            self.LANGUAGE_PARAM not in self.request.GET and
+            self.country_code is not None and
+            self.country_code not in self.DOMESTIC_COUNTRY_CODES
+        )
+
+    def get_response(self):
+        url = '{url}?{param}={lang}'.format(
```
konradko: what if URL already has some query params set, e.g. UTM? This should check if query params are already present and append accordingly. We already do this somewhere with urlparse I believe, FAS or SSO.

PR #410, opened 2018-08-29
[DNM] Add ukef lead generation pardot test form

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-319)

image

DNM. to unit test the feature flag.


conf/urls.py
```diff
@@ -471,6 +467,21 @@
         contact.views.TriageWizardFormView.as_view(),
         name='contact-us-triage-wizard'
     ),
+    url(
+        r"^get-finance/$",
+        finance.views.GetFinanceCMS.as_view(),
+        name='get-finance'
+    ),
+    url(
+        r'^get-finance/start/$',
```
richtier: using a redirect url here instead of putting a hyperlink directly to pardot in the html. the reason is go google links to this page instead of pardot. the reason that is desirable is in future we may change how we track, or change the pardot url, but google or links in emails will still link to the old pardot url.

PR #412, opened 2018-09-05
Add new get finance page

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-348)

# feature flag on

image

# feature flag off

image


finance/views.py
```diff
@@ -49,6 +64,14 @@ def get_context_data(self, *args, **kwargs):
 
 
 class GetFinanceStartRedirectView(FeatureFlagMixin, RedirectView):
+    """
+    Using a redirect URL here instead of putting a hyperlink directly to Pardot
+    in the html. the reason is go Google links to this page instead of Pardot.
```
konradko: whut

finance/views.py
```diff
@@ -49,6 +64,14 @@ def get_context_data(self, *args, **kwargs):
...
+    The reason that is desirable is in future we may change how we track, or
```
konradko: whut

finance/views.py
```diff
@@ -27,13 +27,28 @@ def get_context_data(self, *args, **kwargs):
...
     template_name = 'finance/get_finance.html'
     slug = EXPORT_READINESS_GET_FINANCE_SLUG
 
 
+class GetFinanceNegotiator(TemplateView):
+    def __new__(cls, *args, **kwargs):
+        if settings.FEATURE_FLAGS['UKEF_LEAD_GENERATION_ON']:
+            return GetFinance(*args, **kwargs)
+        else:
+            return DeprecatedGetFinance(*args, **kwargs)
```
konradko: Not a fan of the name, we should have deprecated code around, but this is actually still in use depending on the flag... Not a big deal.

finance/views.py
richtier: as per our convo I turned a github comment into a code comment. will edit the comment to make ti clearer.

finance/views.py
richtier: marking code as deprecated is a step away from deleting it e.g, the deprecation warnings in Django when using the old style middleware.

`DeprecatedGetFinance` is used when the feature flag is turned off, and `DeprecatedGetFinance` will be deleted once the feature is to be turned on always.

PR #414, opened 2018-09-11
Add high potential opportunity views

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-343)

image


invest/templates/invest/high-potential-opportunity-detail.html
```diff
@@ -0,0 +1,231 @@
...
+<section id="other-opportunities">
+    <div class="container">
+        <h2 class="heading-xlarge">{{ page.other_opportunities_title }}</h2>
+        <div class="grid-row">
+            {% for opportunity in page.other_opportunities %}
+                <div class="column-one-half">
+                    <div>
+                        <img src="{{ opportunity.hero_image.url }}" />
+                        <div class="title-container">
+                            <a class="link" href="{% url 'high-potential-opportunity-details' slug=opportunity.meta.slug %}">{{ opportunity.heading }}</a>
```
eadpearce: you can use the cards from directory components here
```
{% url 'high-potential-opportunity-details' slug=opportunity.meta.slug  as opportunity_url %}
{% include 'directory_components/card.html' with img_src=opportunity.hero_image.url url=opportunity_url title=opportunity.heading %}
```
and you can use `card-grid` instead of `grid-row` so they have aligned heights regardless of content 👍 column classes are the same

this needs components v2.22.0

invest/templates/invest/high-potential-opportunity-detail.html
```diff
@@ -0,0 +1,231 @@
...
+                </li>
+            </ul>
+        </div>
+    </div>
+</section>
+
+<section id="case-studies">
+    <div class="container">
+        <div class="grid-row">
+            <h2 class="heading-xlarge">{{ page.case_study_list_title }}</h2>
```
eadpearce: make sure to put headings outside of `grid-row` divs otherwise the outer margins won't line up (grids have a negative margin)

invest/templates/invest/high-potential-opportunity-detail.html
```diff
@@ -0,0 +1,231 @@
...
+                    {{ page.proposition_two_video }}
+                {% else %}
+                    <img src="{{ page.proposition_two_image.url }}" />
+                {% endif %}
+            </div>
+        </div>
+    </div>
+</section>
+
+<section id="competative-advantages">
```
eadpearce: 'competitive' (i'm being pedantic)

invest/templates/invest/high-potential-opportunity-detail.html
```diff
@@ -0,0 +1,231 @@
...
+{% endblock %}
+
+{% block head_title %}
+    High Potential Opportunities - great.gov.uk
+{% endblock %}
+
+{% block content %}
+<section id="hero" style="background-image: url({{ page.hero_image.url }})">
+    <div class="container">
+        <p class="breadcrumbs body-text">
```
eadpearce: components has a breadcrumbs template now https://github.com/uktrade/directory-components/blob/master/directory_components/templates/directory_components/breadcrumbs.html

invest/templates/invest/high-potential-opportunity-detail.html
```diff
@@ -0,0 +1,231 @@
...
+</section>
+
+<section id="proposition-two">
+    <div class="container">
+        <div class="grid-row">
+            <div class="column-one-half">
+                <div class="proposition-two-container">
+                    {{ page.proposition_two|add_export_elements_classes }}
+                </div>
+                <ul class="top-level-list">
```
eadpearce: elements has the classes `panel-vertical-narrow` and `panel-vertical-wide` that add a border to the top in blue. you could use that and override the colour here

invest/templates/invest/high-potential-opportunity-detail.html
```diff
@@ -0,0 +1,231 @@
...
+                </div>
+            </div>
+        </div>
+    </div>
+</section>
+
+<section id="testimonial">
+    <div class="container">
+        <div class="grid-row">
+            {{ page.testimonial|add_export_elements_classes }}
```
eadpearce: this should probably be a `blockquote`

core/views.py
```diff
@@ -171,6 +176,8 @@ def location(self, item):
                 'step': TriageWizardFormView.EXPORTED_BEFORE})
         if item == 'uk-export-finance-lead-generation-form':
             return reverse(item, kwargs={'step': 'contact'})
+        if item in self.high_potential_opportunity_url_names:
```
konradko: Will the tests break if the url names change?

invest/templates/invest/high-potential-opportunities-form.html
```diff
@@ -0,0 +1,25 @@
+{% extends 'core/base.html' %}
+{% load static from staticfiles %}
+
+{% load add_export_elements_classes from directory_components_tags %}
+ {% block head_css %}
+    {{ block.super }}
+    <link href="{% static 'styles/high-potential-opportunities-form.css' %}" rel="stylesheet" type="text/css" />
+{% endblock %}
+ {% block head_title %}
+    High Potential Opportunities - great.gov.uk
```
konradko: should this not be in CMS?

invest/templates/invest/high-potential-opportunities-form.html
```diff
@@ -0,0 +1,25 @@
...
+{% endblock %}
+ {% block content %}
+    <div class="container">
+    	<div id="heading-container">
+	    	<h2 class="heading-large">Your business details</h2>
+	    	<span class="font-xsmall">The questions below will help us provide the best support for you</span>
```
konradko: ditto

invest/templates/invest/high-potential-opportunities-form.html
richtier: yeah, in fact the CMS already supports that field

invest/templates/invest/high-potential-opportunity-detail.html
richtier: had a go, unfortunately, the card component crates one card per row, while we want two cards per row:

### the card
image

### but we want

image

eadpearce:
also, @richtier can you upload a screenshot for the mobile version as well?

invest/templates/invest/high-potential-opportunity-detail.html
richtier: oops :) good eye

invest/templates/invest/high-potential-opportunity-detail.html
```diff
@@ -0,0 +1,231 @@
...
+</section>
+
+<section id="company-list">
+    <div class="container">
+        <div class="grid-row">
+            <div>
+                {{ page.companies_list_text|add_export_elements_classes }}
+            </div>
+            <ul>
+                <li class="company-list-first-line">
```
eadpearce: just looking at mobile design now. these should be in a grid that collapses to 2 columns on mobile. you can use our shiny new responsive grid classes for that :) https://directory-components-dev.herokuapp.com/responsive-grid/

invest/templates/invest/high-potential-opportunity-detail.html
richtier: good stuff :)

food for thought: the breadcrumbs component expects `parent_site` dictionary, which would mean passing `parent_site={'url': '/', 'title': 'Home'}` from the view to the template, which is a bit more complex than it needs to be.

looks like we could update the breadcrumb component to something like
`<a href="{{ root_page_url }}">{{ root_page_breadcrumb_label }}</a>`

richtier: oh oh. the breadcrumb component is using `page.title`. `title` is the cms title :(

I'll use the current html, and tomorrow will contribute to directory components to allow those values to be passed in, rather than pulling them from a dictionary (which is a bit restrictive)

invest/templates/invest/high-potential-opportunity-detail.html
richtier: ### panel-vertical-narrow gives us
image

### but we want
image

that's because the field value is markdown-generated HTML, and here is a list within a list. we want the sub-list to have the borderline thing, but not the parent list.

but good tip - I can use `panel-vertical-narrow` elsewhere:)

invest/templates/invest/high-potential-opportunity-detail.html
richtier: I've not implemented mobile yet. that will be a follow up PR. this is behind a feature flag so it will do no harm, and the PR is getting laaaarge :)

invest/templates/invest/high-potential-opportunity-detail.html
eadpearce: yeah this component needed some tweaking so these are good improvements to add

invest/templates/invest/high-potential-opportunity-detail.html
eadpearce: is that adding the class to the `ul`? couldn't you add it to the `li`s instead? or does that give you some weird alignment issues?

invest/templates/invest/high-potential-opportunity-detail.html
eadpearce: ? no you just gotta put them inside a `column` class. they stretch to fit the size of their container

invest/templates/invest/high-potential-opportunity-detail.html
eadpearce: okay that is cool 👍

invest/templates/invest/high-potential-opportunity-detail.html
eadpearce: there is an example in the components demo page https://github.com/uktrade/directory-components/blob/master/demo/templates/demo/components.html#L108

PR #421, opened 2018-09-14
Add Mobile view for HPO detail page

richtier:
[part of this task](https://uktrade.atlassian.net/browse/TT-420)

### desktop

image

### mobile

image


core/sass/high-potential-opportunity-detail.scss
```diff
@@ -198,12 +197,10 @@
   }
   ul {
     margin-top: 70px;
-    li.company-list-first-line {
+    column-count: 4;
```
eadpearce: `column-count` is not compatible with IE <9. will that be an issue?

richtier: the outcome is acceptable: the items will be shown in a horizontal list

eadpearce: sounds fair 👍

PR #428, opened 2018-09-21
Add templates for missing privacy pages

eadpearce:


core/mixins.py
```diff
@@ -20,8 +20,12 @@ def flag(self):
 
 class GetCMSPageMixin:
     def get_context_data(self, *args, **kwargs):
+        if hasattr(self, 'slug'):
```
richtier: nice flexibility. good stuff.

however, it adds complexity. we can avoid this complexity by expecting the views to define a slug property. either
```
slug = 'some-value'
```
or
```
@property
def slug(self):
     return self.kwargs['slug']
```

simpler :)

eadpearce: the slug is not hardcoded in the view so that if needed content designers can add more subpages without any dev work which I anticipate happening in future

eadpearce: oh i was reading this wrong haha. i can rewrite this

PR #437, opened 2018-09-26
Fix ukef form mobile

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-504)
[this task](https://uktrade.atlassian.net/browse/TT-505)
[this task](https://uktrade.atlassian.net/browse/TT-508)


finance/templates/finance/get_finance.html
```diff
@@ -47,10 +47,10 @@
 
 <section id="advantages-section">
     <div class="container">
+        <div class="column-one-half">
```
eadpearce: why did this move out of the grid-row?

eadpearce: oh i see now. you should put this in its own grid-row then you won't need the clearfix div

PR #450, opened 2018-10-04
Add prototype and news pages

eadpearce:
screencapture-exred-trade-great-8007-news-2018-10-04-16_26_03

screencapture-exred-trade-great-8007-prototype-2018-10-04-16_24_35

screencapture-exred-trade-great-8007-prototype-markets-2018-10-04-16_36_39

screencapture-exred-trade-great-8007-prototype-guidance-market-research-2018-10-04-16_38_10

screencapture-exred-trade-great-8007-prototype-guidance-market-research-define-market-potential-2018-10-04-16_39_14


conf/settings.py
```diff
@@ -404,6 +405,10 @@
 
 # feature flags
 FEATURE_FLAGS = {
+    'PROTOTYPE_PAGES_ENABLED': env.bool(
+        'FEATURE_PROTOTYPE_PAGES_ENABLED', False),
+    'NEWS_SECTION_ENABLED': env.bool(
+        'FEATURE_NEWS_SECTION_ENABLED', False),
```
richtier: sweet feature flags :)

prototype/helpers.py
```diff
@@ -0,0 +1,6 @@
+
+def unprefix_prototype_url(path):
+    prefix = '/prototype'
+    if prefix not in path:
+        return path
+    return path[len(prefix):]
```
richtier: bit more pythonic: 
```
def unprefix_prototype_url(path):
    return path.replace('/prototype', '')
```

prototype/mixins.py
```diff
@@ -0,0 +1,31 @@
...
+from directory_cms_client.client import cms_api_client
+
+from core.helpers import handle_cms_response
+from prototype.helpers import unprefix_prototype_url
+
+
+class CMSPageViewMixin(TemplateView):
+
+    @cached_property
+    def get_cms_page(self):
```
richtier: it's a bit misleading for a property to start with `get_` - the reader would expect that to be called as a function. more accurate would be to call this `page`.

prototype/tests/test_helpers.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from prototype.helpers import unprefix_prototype_url
+
+unprefix_urls = [
+    ('/prototype/markets/', '/markets/'),
+    ('/prototype/guidance/', '/guidance/'),
+    ('/news/', '/news/'),
+    ('/', '/'),
+]
```
richtier: :+1:

prototype/tests/test_views.py
```diff
@@ -0,0 +1,88 @@
...
+
+    mock_get_page.return_value = create_response(
+        status_code=200,
+        json_body=page
+    )
+    response = client.get(url)
+
+    assert response.status_code == 200
+    assert response.template_name == ['prototype/landing_page.html']
+    print(response.context_data)
```
richtier: I suspect this was erroneously committed

prototype/views.py
```diff
@@ -0,0 +1,43 @@
...
+
+class ArticleDetailView(CMSPageViewMixin, SlugFromKwargsMixin):
+    template_name = 'prototype/article_detail.html'
+    url_name = 'article-detail'
+
+    def get_context_data(self, *args, **kwargs):
+
+        social_links_builder = SocialLinkBuilder(
+            self.request.build_absolute_uri(
+                reverse(self.url_name, kwargs={**self.kwargs})
```
richtier: `reverse(self.url_name, kwargs={**self.kwargs})` can be replaced with `self.request.path`

richtier: oh! in fact I think this we can just use `self.request.build_absolute_uri()` without any path provided :)

prototype/mixins.py
```diff
@@ -0,0 +1,31 @@
+from django.views.generic import TemplateView
+from django.utils.functional import cached_property
+
+from directory_cms_client.client import cms_api_client
+
+from core.helpers import handle_cms_response
+from prototype.helpers import unprefix_prototype_url
+
+
+class CMSPageViewMixin(TemplateView):
```
richtier: this is quite similar to core.mixins.GetCMSPageMixin`. it would be nice if we used a similar but different name as that such as `GetCMSPageByUrlPathMixin` or similar. It tells the reader what the mixin does.

conf/urls.py
```diff
@@ -524,3 +528,22 @@
...
+        name='news-article-list',
+    ),
+    url(
+        r"^news/(?P<slug>[\w-]+)/$",
+        prototype.views.NewsArticleDetailView.as_view(),
+        name='news-article-detail',
+    ),
+]
+
+if settings.FEATURE_FLAGS['PROTOTYPE_PAGES_ENABLED']:
```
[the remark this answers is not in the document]
eadpearce: AHA. i thought this might be why the tests don't work haha. yeah i'll do that

core/views.py
```diff
@@ -84,6 +96,18 @@ def get(self, request, *args, **kwargs):
         return super().get(request, *args, **kwargs)
 
 
+class PrototypeLandingPageView(CMSPageViewMixin, LandingPageView):
+    template_name = 'prototype/landing_page.html'
+
+    @cached_property
+    def get_cms_page(self):
```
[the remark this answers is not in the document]
eadpearce: it's used by `CMSPageViewMixin`

prototype/tests/test_views.py
eadpearce: yarp

prototype/views.py
eadpearce: none of those work unfortunately :(

PR #464, opened 2018-10-18
Add international news article pages

eadpearce:
screencapture-exred-trade-great-8007-international-news-2018-10-18-11_54_14
screencapture-exred-trade-great-8007-international-news-news-article-2018-10-18-11_54_28


conf/urls.py
```diff
@@ -538,6 +538,16 @@
         prototype.views.NewsArticleDetailView.as_view(),
         name='news-article-detail',
     ),
+    url(
+        r"^international/news/$",
```
richtier: `international/` or `int/`?

prototype/mixins.py
```diff
@@ -29,3 +31,34 @@ class SlugFromKwargsMixin():
...
     def slug(self):
         return self.kwargs['slug']
+
+
+class InternationalNewsCMSPageMixin(GetCMSPageByFullPathMixin):
+
+    @cached_property
+    def page(self):
+        response = cms_api_client.lookup_by_full_path(
+            full_path=prefix_international_news_url(self.request.path),
```
richtier: you could make this DRYer by adding a property to `GetCMSPageByFullPathMixin` called `cms_lookup_path` that returns `self.request.path`:

```
class GetCMSPageByFullPathMixin(TemplateView):
     @property
     def cms_lookup_path(self):
        return self.request.path

    @cached_property
    def page(self):
        response = cms_api_client.lookup_by_full_path(
            full_path=self.cms_lookup_path,
            draft_token=self.request.GET.get('draft_token'),
        )
        return handle_cms_response(response)
```
then you can do
```
class InternationalNewsCMSPageMixin
    @property
    def cms_lookup_path(self):
        return prefix_international_news_url(self.request.path)
```

that has less duplication of logic :)

prototype/mixins.py
```diff
@@ -29,3 +31,34 @@ class SlugFromKwargsMixin():
...
+            draft_token=self.request.GET.get('draft_token'),
+        )
+        return handle_cms_response(response)
+
+
+class SocialLinksMixin():
+
+    @cached_property
```
richtier: `cached_property` is normally reserved for when the function is performing an "expensive" operation such as retrieving data from an API. It speeds up the code at the tradeoff of adding complexity

this function looks rather cheap, so it's hard to justify the added complexity of using `cached_property`

prototype/templates/prototype/sections/services_section.html
```diff
@@ -1,4 +1,4 @@
-<section id="services" class="services-section background-stone">
+<section id="services" class="services-section background-stone-30">
```
richtier: :heart:

prototype/templatetags/prototype_tags.py
```diff
@@ -8,5 +8,5 @@
 @register.filter
 def parse_date(date_string):
     if date_string:
-        return dateparser.parse(date_string).strftime('%m %B %Y')
+        return dateparser.parse(date_string).strftime('%d %B %Y')
```
richtier: could you add a test to cover this, please

prototype/views.py
```diff
@@ -53,18 +65,22 @@ def get_context_data(self, *args, **kwargs):
...
+):
+    template_name = 'prototype/domestic_news_detail.html'
     url_name = 'news-article-detail'
+
+
+class InternationalNewsArticleDetailView(
+    NewsArticleDetailView, InternationalNewsCMSPageMixin,
+):
+    template_name = 'prototype/international_news_detail.html'
+    url_name = 'international-news-article-detail'
```
richtier: how is url_name used?

core/sass/partials/_effects.scss
```diff
@@ -1,189 +0,0 @@
-// EFFECT STYLES
```
eadpearce: moved this styling into `_carousel` and `_modal` to make it more clear from the naming what it's for

core/sass/partials/_exred-main.scss
```diff
@@ -1,53 +0,0 @@
-// Start Great Pattern Library
```
eadpearce: this styling is no longer used

prototype/mixins.py
eadpearce: ty i was wondering how i could improve that

prototype/mixins.py
eadpearce: okay i wondered whether it was worth it to use that. i can remove it

prototype/views.py
eadpearce: i think it was used in `social_links_builder` but isn't anymore. i can probably remove it

conf/urls.py
eadpearce: international homepage is `international/` so `international/news/` makes sense here. i don't think we use `int/` anywhere

PR #465, opened 2018-10-19
Fix homepage language selector, feature flag banner

eadpearce:


core/views.py
```diff
@@ -113,6 +113,12 @@ class InternationalLandingPageView(
 ):
     template_name = 'core/landing_page_international.html'
 
+    def get_context_data(self, *args, **kwargs):
+        return super().get_context_data(
+            news_section_on=settings.FEATURE_FLAGS['NEWS_SECTION_ON'],
```
richtier: directory components exposes the feature flags to templates

in the template we can do

```
if feature_flags.NEWS_SECTION_ON %}
```

eadpearce: oh neat i didn't realise that

PR #466, opened 2018-10-19
Add tag list pages

eadpearce:
screencapture-exred-trade-great-8007-prototype-tagged-new-to-exporting-2018-10-19-13_43_06


prototype/mixins.py
```diff
@@ -31,6 +31,25 @@ def get_context_data(self, *args, **kwargs):
         )
 
 
+class GetCMSTagMixin(TemplateView):
```
richtier: this does not need to inherit from `TemplateView` - it can be `class GetCMSTagMixin:` :)

prototype/tests/test_views.py
```diff
@@ -299,3 +299,111 @@ def test_news_list_page_feature_flag_off(client, settings):
     response = client.get(url)
 
     assert response.status_code == 404
+
+
+test_articles = [
```
richtier: solid tests :muscle:

prototype/mixins.py
eadpearce: oh fair enough. i can do the same with `GetCMSPageByFullPathMixin`

PR #478, opened 2018-10-26
[DNM] Add events bodw2019 redirect

eadpearce:
do not merge until new events platform is live


richtier:
the wrong header is in use (I think it's wrong...it's certainly different :) )

image

the video is not loading. maybe its

footer is funky 
image

PR #480, opened 2018-10-29
Add banner component to international page

eadpearce:
https://uktrade.atlassian.net/browse/CMS-521


core/mixins.py
```diff
@@ -45,6 +46,29 @@ def get_context_data(self, *args, **kwargs):
...
+            draft_token=self.request.GET.get('draft_token'),
+            service_name=cms.COMPONENTS,
+        )
+        return handle_cms_response_allow_404(response)
+
+    def get_context_data(self, *args, **kwargs):
+
+        component_is_rtl = 'ar' in str(
+            self.cms_component['meta']['languages']
+        ) and translation.get_language() == 'ar'
```
richtier: we can let django do the hard work:
```
from django.utils import translation

activate_language = translation.get_language()
activate_language_is_bidi = translation.get_language_info(activate_language)['bidi']
component_supports_activate_language = activated_language in self.cms_component['meta']['languages']
component_is_rtl = (activate_language_is_bidi and component_supports_activate_language)
```
this will also work for any bidi language, not just ar :)

core/mixins.py
```diff
@@ -45,6 +46,29 @@ def get_context_data(self, *args, **kwargs):
         return super().get_context_data(page=self.page, *args, **kwargs)
 
 
+class GetCMSComponentMixin:
+    @cached_property
+    def cms_component(self):
+        response = cms_api_client.lookup_by_slug(
```
richtier: will we only have a max of 1 component per page?

core/views.py
```diff
@@ -107,9 +107,17 @@ def page(self):
...
+    mixins.GetCMSPageMixin,
+    mixins.GetCMSComponentMixin,
+    TemplateView,
 ):
     template_name = 'core/landing_page_international.html'
+    component_slug = 'eu-exit-banner-international'
+
+    @property
+    def slug(self):
+        return 'international-news'
```
richtier: as this is a static value we could do
```
class InternationalLandingPageView(...):
    slug = 'international-news'
```

core/mixins.py
```diff
@@ -45,6 +46,29 @@ def get_context_data(self, *args, **kwargs):
...
+            slug=self.component_slug,
+            language_code=translation.get_language(),
+            draft_token=self.request.GET.get('draft_token'),
+            service_name=cms.COMPONENTS,
```
richtier: woo! nice change to the client :)

core/templates/core/landing_page_international.html
```diff
@@ -1,11 +1,13 @@
...
 {% load trans from i18n %}
+{% load banner from directory_components_tags %}
 
 {% block content %}
-{% if features.NEWS_SECTION_ON %}
-  <div class="container" dir="ltr">
-    {% include 'core/includes/eu-exit-banner.html' %}
+
+{% if features.NEWS_SECTION_ON and cms_component and page.articles_count > 0 %}
+  <div class="container banner-container" dir="{{ component_is_rtl|yesno:'rtl,ltr' }}">
```
richtier: `component_is_rtl` is slightly misleading because arabic is not right to left: it's bidirectional because it can contain english words within it: so some of the text is RTL and some is LTR...hence django calls it "bidirectional" or "bidi".

so `component_is_bidi` is more accurate :)

core/mixins.py
eadpearce: for the moment yes. if we add more we could return a dictionary of named components instead of just one. i don't know how likely that would be though

core/mixins.py
eadpearce: i forgot this was a thing. i'll change it

core/templates/core/landing_page_international.html
eadpearce: oh sure. i was thinking in terms of what `dir` attribute the html component needed but this makes sense

PR #483, opened 2018-10-29
Send confirmation email of euexit via email

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-663)


conf/settings.py
```diff
@@ -449,3 +449,6 @@
 DIRECTORY_FORMS_API_SENDER_ID_EUEXIT = env.str(
     'DIRECTORY_FORMS_API_SENDER_ID_EUEXIT'
 )
+EUEXIT_AGENT_EMAIL = env.str('EUEXIT_AGENT_EMAIL')
+EUEXIT_GOV_NOTIFY_TEMPLATE_ID = env.str('EUEXIT_GOV_NOTIFY_TEMPLATE_ID')
+EUEXIT_GOV_NOTIFY_REPLY_TO_ID = env.str('EUEXIT_GOV_NOTIFY_REPLY_TO_ID', None)
```
kowalcj0: Should this default to `None`?

richtier: si. this results in the default "reply to" email to be used

PR #484, opened 2018-10-29
Add css prune task, remove unused styles

eadpearce:


richtier:
@eadpearce any chance we can stop snyk complaining about `mem`?

https://app.snyk.io/vuln/npm:mem:20180117#remediation

realistically this package is used i our build env, not exposed to the internet so the risk is nil, but if enquieting snyk is easy then :+1:

PR #489, opened 2018-10-31
Update tag pages

eadpearce:


makefile
```diff
@@ -187,7 +187,9 @@ DEBUG_SET_ENV_VARS := \
 	export EU_EXIT_ZENDESK_SUBDOMAIN=debug; \
 	export EUEXIT_AGENT_EMAIL=test@example.com; \
 	export EUEXIT_GOV_NOTIFY_TEMPLATE_ID=debug; \
-	export EUEXIT_GOV_NOTIFY_REPLY_TO_ID=debug
+	export EUEXIT_GOV_NOTIFY_REPLY_TO_ID=debug; \
+	export DIRECTORY_FORMS_API_API_KEY_EUEXIT=debug; \
+	export DIRECTORY_FORMS_API_SENDER_ID_EUEXIT=debug
```
richtier: :'( 

hard-coding this kills local dev env integration with forms API because there no sane default for this value (the value is from a database record on locally running Forms API and so will be different on each host).

please remove this and add a `conf/.env` file (which is uncommitted) that can contain:

```
DIRECTORY_FORMS_API_API_KEY_EUEXIT=debug
DIRECTORY_FORMS_API_SENDER_ID_EUEXIT=debug
```

(mine contains a valid api key and sender id that works for my local machine)

PR #490, opened 2018-10-31
Fix error when cms component not found

eadpearce:


core/mixins.py
```diff
@@ -63,14 +63,19 @@ def get_context_data(self, *args, **kwargs):
...
             activated_language)['bidi']
 
-        component_supports_activated_language = activated_language in \
-            self.cms_component['meta']['languages']
-        component_is_bidi = activated_language_is_bidi and \
-            component_supports_activated_language
+        cms_component = None
+        component_is_bidi = activated_language_is_bidi
+
+        if self.cms_component:
```
richtier: this would be more readable if we moved this to it's own property `is_cms_component_bidi`:

```
@property
def is_cms_component_bidi(self):
    if not self.cms_component:
        return False
    component_supports_activated_language = activated_language in \
  ...

PR #506, opened 2018-11-12
International contact page

eadpearce:
https://uktrade.atlassian.net/browse/CMS-599

screencapture-exred-trade-great-8007-international-contact-us-2018-11-12-10_43_32


core/helpers.py
```diff
@@ -17,6 +18,18 @@ def build_social_link(template, request, title):
     )
 
 
+def cms_component_supports_activated_language(activated_language, languages):
+    for language in languages:
```
richtier: we can unpack the `languages` list like so:

```
    for language_code, label in languages:
        if language_code == activated_language:
            return True

```
when unpacking a nice convention is to use `_` for the "thrown away" variable, and we can, and we can use `any` for terseness:

```
return any(code == activated_language for code, _ in languages)
```
quite pretty :)

core/helpers.py
```diff
@@ -17,6 +18,18 @@ def build_social_link(template, request, title):
...
+        if language[0] == activated_language:
+            return True
+
+
+def cms_component_is_bidi(activated_language, languages):
+    if cms_component_supports_activated_language(
+     activated_language, languages):
```
richtier: the indentation is a bit funky

eadpearce: had to add an extra space as linter was complaining. is there a better way to do this?

core/helpers.py
eadpearce: whaaaaaaaat that's neat. i love python

core/tests/test_views.py
```diff
@@ -253,12 +253,27 @@ def test_privacy_cookies_cms(
...
+    )
+)
 @patch('core.views.InternationalLandingPageView.cms_component',
        new_callable=PropertyMock)
 @patch('core.views.InternationalLandingPageView.page',
        new_callable=PropertyMock)
 def test_international_landing_page_news_section_on(
-    mock_get_page, mock_get_component, client, settings
+    mock_get_page, mock_get_component, activated_language,
+    component_languages, dir, client, settings
```
richtier: can you please use a different variable name for `dir` - `dir` is actually a [built-in function](https://www.programiz.com/python-programming/methods/built-in/dir)

core/helpers.py
richtier: :heart:

core/helpers.py
richtier: could do 
```
is_supported = cms_component_supports_activated_language(
    activated_language, languages
)
if is_supported:
    return translation.get_language_info(activated_language)['bidi']
```

buuut now that `cms_component_supports_activated_language` is one line we could simple do:
```
def cms_component_is_bidi(activated_language, languages):
    if any(code == activated_language for code, _ in languages):
        return translation.get_language_info(activated_language)['bidi']
```

eadpearce: nvm i think i fixed it

eadpearce: hmmm okay maybe i'll do that instead

PR #512, opened 2018-11-15
Feature flag contact link on international footer

richtier:
## disabled

image

## enabled
image


core/mixins.py
```diff
@@ -59,19 +59,21 @@ def cms_component(self):
         )
         return handle_cms_response_allow_404(response)
 
-    def get_context_data(self, *args, **kwargs):
-        cms_component = None
-
+    @property
+    def component_is_bidi(self):
```
richtier: I refactored this because of `cms_component` is `None` the  `component_is_bidi` would be undefined

PR #521, opened 2018-11-20
Expose form ingress url

richtier:
results in this:

image


core/sass/contact.scss
```diff
@@ -39,7 +39,7 @@ button {
   }
 }
 
-.your-details-container::before {
+#id_given_name-container::before {
```
kowalcj0: :+1:

PR #529, opened 2018-11-22
Fix back button on contact routing form

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-760)


contact/views.py
```diff
@@ -196,6 +205,11 @@ def render_next_step(self, form):
             return redirect(redirect_url)
         return self.render_goto_step(choice)
 
+    def get_prev_step(self, step=None):
+        if step is None:
```
sdonk: can step be any other falsy value? in that case `if not step` would cover it

PR #535, opened 2018-11-26
Add marketing campaign page

eadpearce:
screencapture-exred-trade-great-8007-prototype-campaign-2018-11-19-18_05_01


marketing/views.py
```diff
@@ -0,0 +1,14 @@
+from django.views.generic import TemplateView
+
+from core.mixins import GetCMSPageMixin
+
+
+class CampaignPageView(
+    GetCMSPageMixin,
+    TemplateView
```
richtier: these could all be one line - would be easier to read as `class CampaignPageView(GetCMSPageMixin, TemplateView)`

richtier: I suspect this needs to be feature flagged

conf/settings.py
```diff
@@ -68,6 +68,7 @@
     'export_elements',
     'directory_components',
     'prototype',
+    'marketing',
     'euexit',
```
richtier: can we be more specific - technically everything in our services is marketing (we're a wholly-owned subsidiary of the marketing department, after all)

richtier: naming is hard

richtier: as discucced - can you please move these to `core`

core/mixins.py
```diff
@@ -16,6 +16,12 @@ def dispatch(self, *args, **kwargs):
         return super().dispatch(*args, **kwargs)
 
 
+class CampaignPagesFeatureFlagMixin(NotFoundOnDisabledFeature):
+    @property
+    def flag(self):
+        return settings.FEATURE_FLAGS['CAMPAIGN_PAGES_ON']
```
richtier: I was going to suggest that as this is not a dynamic value we can do:

```
class CampaignPagesFeatureFlagMixin(NotFoundOnDisabledFeature):
    fag = settings.FEATURE_FLAGS['CAMPAIGN_PAGES_ON']
```

**but** your way is easier to test :+1:

eadpearce: don't look at me i just copypasted one of the other feature flag mixins :'D

PR #549, opened 2018-12-04
Fix terms and conditions click on contact form

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-810)


core/sass/partials/_forms.scss
```diff
@@ -19,6 +19,9 @@
 // @mixin body-s
 // @mixin body-m
 
+#content .form-group input[type=checkbox] {
```
eadpearce: can you move this into `contact.scss`. seems to belong with the other styling for the contact forms rather than generic form styling

PR #555, opened 2018-12-13
Remove "if something is wrong"

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-867)

### DIFF
--- a/contact/forms.py
+++ b/contact/forms.py
@@ -204,9 +204,7 @@ class FeedbackForm(SerializeDataMixin, ZendeskActionMixin, forms.Form):
 
 class BaseShortForm(forms.Form):
     comment = fields.CharField(
-        label=(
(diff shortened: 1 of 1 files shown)


contact/forms.py
```diff
@@ -204,9 +204,7 @@ def full_name(self):
 
 class BaseShortForm(forms.Form):
     comment = fields.CharField(
-        label=(
-            'If something is wrong, please give as much details as you can'
-        ),
+        label='Please give as much details as you can',
```
eadpearce: it should either be "give as much detail" or "give as many details". if it's just a typo use the first one but check with someone from content as well

richtier: true. funny thing is content asked us to say "details" :laughing: 
will raise with them

PR #579, opened 2019-01-03
Update readme

depsiatwal:
updated readme.md with instructions for Redis install and setting environment variables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ We use SASS CSS pre-compiler. If you're doing front-end work your local machine 
     $ redis-server
 warning: This installs the latest version of Redis server on local machine not the pinned version.
 
-### Setting enviroment variables for running the webserver
+### Setting enviroment variables for running the web server
   Locate from the repo file 'makefile' and within the bash prompt run all the export variables from 'DEBUG_SET_ENV_VARS'  &  'TEST_SET_ENV_VARS'
 
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![codecov-image]][codecov]
 [![snyk-image]][snyk]
 
-**Export Readiness (Exred) - the Department for International Trade (DIT)**  
+**Export Readiness (Exred) - the Department for International Trade (DIT)**
 
 ---
 ### See also:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,14 +46,23 @@ We use SASS CSS pre-compiler. If you're doing front-end work your local machine
 ### Installing
     $ git clone https://github.com/uktrade/directory-ui-export-readiness
     $ cd directory-ui-export-readiness
-    $ virtualenv .venv -p python3.5
+    $ virtualenv .venv -p python3.6
     $ source .venv/bin/activate
-    $ pip install -r requirements_test.txt
+    $ python -m  pip install -r requirements_test.txt
```
richtier: did `pip install -r requirements_test.txt` not work for you?

the standard way to run `pip` is to call it directly

README.md
```diff
@@ -46,14 +46,23 @@ We use SASS CSS pre-compiler. If you're doing front-end work your local machine
...
 
 ### Running the webserver
     $ source .venv/bin/activate
     $ make debug_webserver
 
+### Running Redis Server
+    $ brew install redis
```
richtier: Instead of including platform-specific instructions for installing tools we can get away with simply saying something "Install redis on your host"

`brew` assumes mac is being used. Some devs use linux. We don't want the readme to be a general technical manual :)

README.md
```diff
@@ -46,14 +46,23 @@ We use SASS CSS pre-compiler. If you're doing front-end work your local machine
...
+    $ redis-server
+warning: This installs the latest version of Redis server on local machine not the pinned version.
```
richtier: Please update to state version 3.2 should be used. That's what is used in prod and stage.

README.md
```diff
@@ -46,14 +46,23 @@ We use SASS CSS pre-compiler. If you're doing front-end work your local machine
...
+
+### Setting enviroment variables for running the web server
+  Locate from the repo file 'makefile' and within the bash prompt run all the export variables from 'DEBUG_SET_ENV_VARS'  &  'TEST_SET_ENV_VARS'
+
```
richtier: this will cause clashes between env vars on different services. e.g, SSO, API, and CMS all have an env var called `DATABASE_URL`

Instead call the `Makefile` commands such as. This will handle settings the env vars for us without the clashing occuring:
* `make debug_webserver` - runs the django webserver
* `make debug_test` - run the tests
* make debug_manage cmd="<command>" - run an arbitrary management command

Feel free to update the readme to suggest that the dev becomes familar with the `Makefile` commands

README.md
depsiatwal: Issue with that was that the machine was attempting to use default python version on the machine 2.7 even after having an active environment so failed for me. Happy to revert if you think former is more standard

richtier:
reopen when ready to update readme

PR #588, opened 2019-01-14
Update performance dashboard template to fit cms changes

eadpearce:
https://uktrade.atlassian.net/browse/CMS-710


core/templates/core/performance_dashboard.html
```diff
@@ -28,7 +28,7 @@
       <li>
         <a href="{% url 'performance-dashboard' %}">Performance dashboard</a>
       </li>
-      <li aria-current="page">{{ page.title }}</li>
+      <li aria-current="page">{{ page.heading }}</li>
```
richtier: :heart:

PR #598, opened 2019-01-22
Infer company is not in companies house

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-937)

none-of-these


core/sass/components/_form-wizard.scss
```diff
@@ -85,6 +85,10 @@
       color: white;
     }
   }
+  #triage-company-name-not-in-companies-house {
```
eadpearce: since we removed the original triage can we change these to something like `contact-triage-` instead? just to be explicit (doesn't have to be in this PR if there are a lot of them to change)

PR #609, opened 2019-02-07
Add EU exit banner to home page

richtier:
[part of this task](https://uktrade.atlassian.net/browse/TT-1049)


core/tests/test_views.py
```diff
@@ -21,6 +21,8 @@
 @patch('directory_cms_client.client.cms_api_client.lookup_by_slug')
 def test_landing_page_video_url(mock_get_page, client, settings):
     settings.FEATURE_FLAGS['NEWS_SECTION_ON'] = False
+    settings.FEATURE_FLAGS['LANDING_PAGE_EU_EXIT_BANNER_ON'] = False
```
eadpearce: can you make a separate test that checks the template with the feature flag on/off

article/templates/article/landing_page.html
```diff
@@ -47,6 +48,13 @@ <h1 class="heading-xlarge heading-hero-domestic title" id="hero-campaign-section
   </div>
 </section>
 
+{% if features.LANDING_PAGE_EU_EXIT_BANNER_ON %}
+  <br>
```
eadpearce: ideally shouldn't use `br` tags for spacing unless it's in a block of text. if you need more space at the top of the banner you can add a class like `padding-top-30` or `margin-top-30`

PR #610, opened 2019-02-07
Improve style of EU exit banner

richtier:
image


article/templates/article/landing_page.html
```diff
@@ -50,8 +49,10 @@ <h1 class="heading-xlarge heading-hero-domestic title" id="hero-campaign-section
 
 {% if features.LANDING_PAGE_EU_EXIT_BANNER_ON %}
   <br>
-  <div class="container">
-    {% banner banner_content=page.banner_content badge_content=page.banner_label %}
+  <div class="container eu-exit-banner">
+     <div>
+       <strong>{{ page.banner_label }}</strong><span>{{ page.banner_content|safe }}</span>
```
eadpearce: semantically i think the `banner_label` should be an `h2` and the `banner_content` should be a `p`. just style them so they're both `display: inline` and that should work

